Issue #53, opened 2026-07-29
In-body <a href> never becomes a KFX link — footnote markers are not tappable


jloutsch:
## Symptom

Footnote/endnote markers in generated KFX are not tappable — the number sits at the
end of the sentence as plain text and does not jump to the note. The reverse link
(note → body) is missing too.

## Root cause

`<a href>` inside body text is never turned into a link.

`converter.py::_walk_inline` has no case for `<a>`: the element is flattened to its
text and the `href` is discarded. `<a>` is read only by `_own_anchor_ids` /
`_subtree_anchor_ids` (converter.py:149-167), which harvest **ids**, and those ids
feed only `_anchor_block_index` (converter.py:426) — used for chapter splitting.
No `$266` anchor fragment is ever emitted for a harvested id.

So both halves are missing: no source link span on the marker, and no target anchor
at the note.

The only link-emitting path in the codebase is the synthesized Contents page:
`_rebuild_contents_page` → `toc_links` → `native_generator` (converter.py:810).
In a representative 975-noteref book the generated KFX contains exactly 24 `$179`
link refs, all `toc_anchor_*`, and 24 `$266` anchors — one per contents entry, none
for the 975 note targets.

## Fix

1. Capture `<a href>` runs in `_walk_inline` as `(start, length, href)` alongside the
   existing bold/italic flags.
2. Emit a `$266` anchor fragment for each harvested `anchor_id` that is actually
   referenced by some href, pointing at the containing block's eid.
3. Resolve `file.xhtml#frag` (and bare `#frag`) to the anchor name at generation time;
   drop external schemes and unresolvable targets rather than emitting dangling `$179`.
4. Reuse the existing span-level `$142` + `$179` shape, which is already correct.

Cross-file resolution matters here: note markers live in chapter files and point into
`endnotes.xhtml`, and each note points back into its chapter.

## Depends on

#51 — anchors need `$180` before any link can resolve, including the ones this issue adds.

jloutsch:
Shipped in **v5.6.0** and device-verified: tapping a footnote marker jumps to the notes section on a physical Kindle.

In-body `<a href>` becomes a `$142` span carrying `$179`, with targets resolved across spine files and anchors emitted for referenced ids. External and unresolvable targets are dropped rather than left dangling. Extended in v5.6.1 by #62 (back-matter and whole-file targets) and #69 (directory-aware keys).

Left open after merge because the squash from a stacked branch did not trigger auto-close.

Issue #52, opened 2026-07-29
Superscript (<sup> and CSS vertical-align) is dropped — footnote markers render at baseline


jloutsch:
## Symptom

Endnote/footnote reference numbers render as full-size baseline digits in the
generated KFX. In the source EPUB they are superscripts.

## Root cause

kfxgen has no inline model for superscript.

`plugin/kfxgen/inline_style.py` defines exactly two inline flags, `FLAG_ITALIC` and
`FLAG_BOLD`. `converter.py::_walk_inline` accumulates them from `<em>/<i>/<strong>/<b>`
only, so a literal `<sup>` is flattened to plain text.

The more common case is worse: publisher EPUBs usually get superscript from CSS, not
markup. A representative sample has 975 note references shaped like

```html
<span class="EN_REF"><a epub:type="noteref" href="endnotes.xhtml#note23">1</a></span>
```

```css
span.EN_REF { vertical-align: super; font-size: 80%; }
```

with no `<sup>` anywhere in the body text. `compute_block_style` never reads
`vertical-align`, and it is block-level anyway, so the span is invisible to it.

## How real KFX encodes superscript

`$31` is baseline shift. Amazon-produced noteref character styles:

```
reference B  s1XR {$16: 0.75 rem, $42: 1.33249, $31: {1.8, $318(pt)}}
reference C  s33N {$16: 0.75 rem, $42: 1.33249, $31: {35,  $314(%)}}
```

Superscript is reduced `$16` font-size plus a positive `$31`, not a dedicated flag.
In reference C exactly 2 styles carry `$31` and both are noteref styles.
kfxgen emits `$31` nowhere.

## Fix

Add a superscript/subscript inline flag sourced from **both** `<sup>`/`<sub>` tags and
computed `vertical-align: super | sub | <length>`, and emit it as a `$142` character
span whose `$157` style carries `$16` (~0.75rem) and `$31` (positive for super,
negative for sub).

Note `$31` accepts either `$318` (pt) or `$314` (%) units in reference files.

## Related

Blocks nothing, but pairs with the footnote-linking work — the same character run is
both a superscript and a link.

jloutsch:
Shipped in **v5.6.0** and device-verified: endnote reference numbers render as superscripts on a physical Kindle.

[...]

PR #54, opened 2026-07-29
fix: working footnotes — anchors, superscript, in-body links (#51, #52, #53)

### LINKED ISSUE #51: KFX $266 anchors omit the $180 self-name, so all internal links are dead
## Symptom

Internal links in generated KFX do not work. On a physical Kindle the synthesized
Contents page renders each chapter title as a plain line rather than a tappable link.

## Root cause

`$266` anchor fragments are missing the `$180` self-name field.

Comparing kfxgen output against four local reference KFX files (one produced by
Calibre's KFX Output plugin, three produced by Amazon):

```
kfxgen     $266 fid=toc_anchor_0  {$183: {$155: 1006, $143: 0}}
reference  $266 fid=a2DK          {$180: a2DK, $183: {$155: 2451}}
reference  $266 fid=a1UE          {$180: a1UE, $183: {$155: 1678}}
```

`$266` shape counts across the reference files — every anchor is either
`($180, $183)` or `($180, $186)`. 100% carry `$180`; none carry `$143` inside `$183`:

| reference | ($180,$183) | ($180,$186) |
|---|---|---|
| A (Calibre KFX Output) | 137 | 74 |
| B (Amazon) | 676 | 126 |
| C (Amazon) | 1015 | 11 |
| D (Amazon) | 3228 | 47 |

The link *spans* kfxgen emits are already correct and match the references exactly:

```
{$155: 1070, $157: s1, $142: [{$143: 0, $144: 16, $179: toc_anchor_0, $157: s0_link}]}
```

So `$179` resolves an anchor by name, and the anchor never declares one.

kfxgen already does the equivalent self-naming for styles (`$173: s0` on every `$157`),
which suggests this is an oversight rather than a deliberate choice.

## Fix

`plugin/kfxgen/native_generator.py` `build_fragment_266`: add `$180: <anchor_name>`,
drop the `$143: 0`.

## Verification

Cannot be rendered locally — raw `.kfx` is rejected by Previewer, so confirmation
needs a sideload to physical hardware. Structural assertion against the reference
shape is the automated check.

jloutsch:
Fixes #51, #52, #53.

Footnote markers rendered as plain baseline digits that went nowhere, and the
Contents page rendered chapter titles as plain lines. Investigating turned up
three separate defects, not one — found by diffing generated output against one
Calibre KFX Output build and three Amazon-produced KFX files (all local,
gitignored, not named here).

## #51 — anchors had no name

The link *spans* were already correct and matched the references exactly. The
`$266` anchors they pointed at were not: an anchor's name lives in `$180`, and
`build_fragment_266` never emitted it, so `$179` resolved against nothing and
every internal link in every book was dead.

```
before   $266 fid=toc_anchor_0  {$183: {$155: 1006, $143: 0}}
after    $266 fid=toc_anchor_0  {$180: toc_anchor_0, $183: {$155: 1006}}
```

100% of anchors in all four reference files carry `$180`; none carries the
`$143` we were adding inside `$183`.

## #52 — superscript was not modelled

The inline model had exactly two flags, bold and italic. `<sup>`/`<sub>`
flattened to text. Worse for real books: publisher EPUBs get superscript from
CSS, not markup — the marker is a `<span>` whose class carries `vertical-align`,
with no `<sup>` anywhere in the body.

Both routes are now recognized, including the raw-length form
(`vertical-align: 0.25em`) publishers use as often as the `super` keyword. A
raised run becomes a `$157` character style with reduced `$16` font-size plus a
`$31` baseline shift, matching the reference noteref styles.

## #53 — in-body `<a href>` never became a link

The element was flattened to its text and the href discarded, and no `$266`
anchor was ever emitted for the ids the notes declare — both halves missing.
Link runs are now captured with targets resolved across spine files, and anchors
are emitted only for ids something actually links to (emitting one per id in the
book would be thousands of fragments for no reading benefit). External and
unresolvable targets are dropped rather than left dangling.

## Result on a 975-noteref book

| | before | after |
|---|---|---|
| `$179` link spans | 24 (all Contents) | 4654 |
| dangling targets | — | 0 |
| `$266` anchors | 24 | 2388 |

2388 is in the same range as the Amazon references (802–3228). Generation runs in
0.14s and output is 1409 KB — essentially unchanged.

## Verification limits

- Raw `.kfx` can't be rendered locally, so on-device behavior needs a physical
  sideload. Structural assertions against the reference shapes are the automated
  check.
- The CSS `vertical-align` route depends on Calibre's Stylizer, which only exists
  inside Calibre. The local end-to-end run uses the test shim, so it exercises
  only the `<sup>` tag route. The resolver contract is unit-tested with an
  injected stylizer, consistent with how the other CSS properties are covered.
- Subscript depth has no reference to copy (none of the four files uses one); it
  is the superscript mechanism with a negative shift and is **not**
  device-verified.

## Tests

30 new tests, written failing first. Full suite: 560 passed, 12 skipped.
`ruff check` and `ruff format --check` both clean.

Version bumped to 5.6.0 — the `v5.6.0` tag still needs pushing on merge so the
release publishes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,53 @@
 # Changelog
 
+## 5.6.0 — Working footnotes: anchors, superscript, in-body links (#51, #52, #53)
+
+Footnote markers now render as superscripts and jump to their notes, and the
+Contents page links work. Three separate defects, found by diffing a generated
+book against one Calibre KFX Output build and three Amazon-produced KFX files.
+
+**Fixed (#51):** every internal link was dead, including the synthesized Contents
+page — chapter titles rendered as plain lines. The link *spans* were already
+correct; the `$266` anchors they pointed at were not. An anchor's name lives in
+`$180`, and `build_fragment_266` never emitted it, so `$179` resolved against
+nothing. Every anchor in all four reference files is `($180, $183)` or
+`($180, $186)` — 100% carry `$180`, and none carry the `$143` kfxgen was adding
+inside `$183`. Anchors now match that shape. (kfxgen already self-named styles
+this way — `$173` on every `$157` — so this was an oversight, not a design.)
+
+**Fixed (#52):** superscript was dropped entirely, so endnote reference numbers
+sat on the baseline at full size. The inline model had exactly two flags, bold and
+italic; `<sup>`/`<sub>` were flattened to text. Worse for real books: publisher
+EPUBs get superscript from CSS, not markup — a marker is a `<span>` whose class
+carries `vertical-align`, with no `<sup>` anywhere. Both routes are now
+recognized, including the raw-length form (`vertical-align: 0.25em`) publishers
+use as often as the `super` keyword. A raised run becomes a `$157` character style
+with reduced `$16` font-size plus a `$31` baseline shift, matching the reference
+noteref styles. Subscript is the same mechanism with a negative shift; unlike the
+superscript value it has no reference to copy and is not device-verified.
+
+**Fixed (#53):** `<a href>` in body text never became a link — the element was
+flattened to its text and the href discarded, and no `$266` anchor was ever
+emitted for the ids the notes declared. Both halves of a footnote link were
+missing. Link runs are now captured with their targets resolved across spine
+files (`endnotes.xhtml#note23` from a chapter, `#frag` against the file it appears
+in), and anchors are emitted for the ids something actually links to — not for
+every id in the book, which would be thousands of fragments for no reading
+benefit. Unresolvable and external targets (`http:`, `mailto:`, traversal,
+absolute paths) are dropped rather than emitted as dangling references.
+
+On a 975-noteref book this takes the output from 24 link spans (all Contents) to
+4654 with zero dangling targets, and from 24 anchors to 2388 — in the same range
+as the Amazon reference files (802–3228). Generation time and file size are
+essentially unchanged.
+
+**Verification limits:** raw `.kfx` can't be rendered locally, so on-device
+behavior needs a physical sideload. The CSS `vertical-align` route also depends on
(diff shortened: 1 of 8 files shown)


jloutsch:
Closing one of the stated verification gaps.

Ran a real `ebook-convert` through Calibre 8 with this branch built and installed into an **isolated `CALIBRE_CONFIG_DIRECTORY`** (the real Calibre config was left on 5.5.0 and untouched). This exercises the CSS `vertical-align` path via Calibre's Stylizer, which the test shim cannot reach.

```
fragments:                    2522
$266 anchors:                 2365   (all carry $180)
$179 link spans:              4604   dangling: 0
$157 styles with $31:            1   s1_em: $31=35%, $16=0.75
runs rendered as superscript: 1009
```

The 1009 superscript runs are the CSS-driven noterefs — the shim run only reached 4 (the literal `<sup>` tags). So #52 is confirmed working through the real plugin path, not just at the unit level.

Still outstanding: on-device rendering. Raw `.kfx` is rejected by Previewer, so a physical sideload remains the only way to confirm the markers actually appear raised and the taps actually jump. Subscript is still unverified against any reference.

jloutsch:
## Device-verified ✅

Sideloaded to a physical Kindle. All three fixes confirmed working on device:

- **#51** — the Contents page has working links again.
- **#52** — endnote reference numbers render as actual superscripts.
- **#53** — tapping a marker jumps to the notes section.

This closes the last open caveat on this PR. The `$180` anchor fix was the leading
hypothesis rather than a proven cause when this was opened — since the links now
fire on hardware, it is confirmed: a `$266` anchor with no `$180` cannot be
resolved, and that is what made every internal link dead.

Remaining unverified: **subscript only**. No reference file in the corpus uses one,
so its `$31` depth (-20%) is still a reasoned guess rather than a copied value, and
nothing in the test book exercises it.

PR #57, opened 2026-07-29
fix: gate image optimization on byte size, not just dimensions (#55)

### LINKED ISSUE #55: Image optimization never gates on byte size — near-lossless images pass through untouched (7.7 MB in one book)
## Symptom that surfaced it

A converted book's cover does not display on the Kindle. The cover is structurally
fine — correct `$164` descriptor, correct `resource/cover_img` payload, correct
`cover_image` key in `kindle_title_metadata`, and the JPEG decodes cleanly
(baseline, 3-component RGB, EXIF orientation=1, valid EOI).

What stands out is its size: **2,254,253 bytes** at 1161x1800.

For comparison, covers in books that render correctly:

| source | dimensions | bytes |
|---|---|---|
| this book (cover missing) | 1161x1800 | **2,254,253** |
| another kfxgen book (renders) | 910x1400 | 127,904 |
| Calibre KFX Output reference | 1414x2200 | 237,244 |
| Amazon reference | 1423x2200 | 465,635 |

It also carries ~20 KB of publisher metadata none of the working covers have:
EXIF (5,479) + Photoshop IRB APP13 (7,320) + a second EXIF (4,828) + ICC (3,160)
+ Adobe APP14.

## The defect

`optimize_image` gates purely on dimensions:

```python
size = _read_image_size(data)
if size is None:
    return data
if max(size) <= max_dim:      # DEFAULT_MAX_DIM = 2048
    return data
```

`max(1161, 1800) = 1800 <= 2048`, so the image is returned unchanged. There is no
byte-size condition anywhere in the function. An image with modest dimensions but
near-lossless encoding sails straight through.

This is not one unlucky image. **Every image in that book is under 2048px, so not
one of them was optimized** — 7,677,756 bytes passed through untouched:

```
resource/cover_img        2,254,253  1161x1800   SKIP
resource/img_7            1,177,453  1356x1725   SKIP
resource/img_0            1,129,397  1351x880    SKIP
resource/img_10             729,365  1365x1535   SKIP
resource/img_3              682,323  1357x871    SKIP
...16 of 16 skipped
```

Re-encoding the cover at q85 **at its original dimensions** via the same
`calibre.utils.img.scale_image` call the function already makes yields 490,594
bytes — 21.8% of the original, landing right in the range of the working
references.

## Proposed fix

Add a byte-size trigger alongside the dimension trigger: optimize when
`max(size) > max_dim` **or** `len(data) > max_bytes`, re-encoding at existing
dimensions in the latter case. Env-overridable like the existing knobs
(`KFXGEN_IMAGE_MAX_DIM`, `KFXGEN_IMAGE_QUALITY`), and keep the existing
"never return something larger than the input" guard.

Threshold needs picking against the reference corpus rather than guessed — the
working examples cluster at 128 KB–466 KB.

## Not yet established

That byte size is *why* the Kindle refuses this cover. That is the leading
hypothesis, not a proven one — the competing explanation is the unusual
EXIF/Photoshop/ICC/Adobe header stack confusing the decoder. Both are fixed by
re-encoding, so the fix is the same either way, but the diagnosis should be
confirmed.

An A/B pair has been built to settle it: two KFX files identical in every respect
except cover encoding (2,254,253 vs 490,594 bytes, same pixel dimensions, same
`cover_image` metadata). Sideloading both distinguishes the two hypotheses and
confirms whether the cover is the variable at all.

Note the source EPUB for the affected book is not available locally, so the
per-book pipeline behavior could not be re-run — the analysis is from the
generated `.kfx` plus a reconstruction.

jloutsch:
Fixes #55.

## What the source EPUB settled

With the EPUB in hand, two things are now proven rather than inferred:

- **kfxgen picked the right cover.** The EPUB declares `<meta name="cover" content="cover-image"/>` → the cover image file, and that is exactly what was embedded. The similarly-sized `img_6` (1168x1800 vs the cover's 1161x1800) is the title-page art, correctly kept separate. **The cover-selection hypothesis is dead.**
- **The cover is embedded byte-for-byte verbatim** — identical SHA-256 to the EPUB source. Pass-through is confirmed, not deduced from the gate arithmetic.

## The defect

```python
if max(size) <= max_dim:   # DEFAULT_MAX_DIM = 2048
    return data
```

`max(1161, 1800) = 1800 ≤ 2048`, so a 2.25 MB image is returned untouched. And since **every** image in that book is under 2048px, none of its 16 images was optimized — 7,677,756 bytes embedded verbatim.

## The fix

A second, independent trigger on encoded weight. When only the byte trigger fires, the image is re-encoded **at its existing dimensions** — it is too heavy, not too big. The existing never-return-something-larger guard still applies.

Default 1 MiB, override via `KFXGEN_IMAGE_MAX_BYTES`.

## Choosing the default from data, not taste

My first measurement was **wrong and I corrected it**: I included two files with >1 MB images as "references", then checked their provenance — both were kfxgen's own output (resource id `cover_img`; Amazon and KFX Output use `eN`/`rsrcN`). One of them, despite being named `test_calibre_output.kfx`, has a cover of *exactly* 1,048,576 bytes ending in `0000` instead of `ffd9` — a JPEG truncated at exactly 2^20. That is a February artifact; no 1 MiB constant exists in the current code, so I have not chased it, but it is worth knowing about.

Genuine references only — 212 images across three Amazon-produced KFX files:

| | |
|---|---|
| largest single image | **465,635 bytes** |
| images over 512 KB | **0** |
| median | 43,409 |

1 MiB is therefore ~2x the observed ceiling: it catches the pathological tail without recompressing merely-large images. `KFXGEN_IMAGE_MAX_BYTES=512000` tracks Amazon more tightly.

## Result on the real book

```
total images  7,677,756 -> 4,350,071 bytes  (43% smaller, no dimension changes)
cover         2,254,253 ->   490,594 bytes  (still 1161x1800)
images >1MiB          3 -> 0
```

## What this does NOT claim

That it fixes "cover does not display on Kindle". That report surfaced the defect and the sizes are conspicuous, but the causal link is **unconfirmed** — the competing explanation is the EXIF + Photoshop + ICC + Adobe header stack, which none of the working covers carry. Both are resolved by re-encoding, so the fix is identical either way, but the diagnosis is not established and the changelog says so.

A controlled A/B is built and awaiting a sideload: same plugin build, same book, differing **only** by `KFXGEN_IMAGE_MAX_BYTES` — so cover encoding is the single variable. (My first attempt at this pair was confounded — the two files came from different plugin builds — and was rebuilt.)

## Version

No version bump here, deliberately: #54 already claims 5.6.0 and bumping in both guarantees a conflict. The changelog entry is under `Unreleased` (a new heading for this repo) to be folded into whichever release lands first.

## Tests

7 new tests, written failing first, covering both triggers, dimension preservation on the byte path, env override, and the never-grow guard. One pre-existing test needed its spy signature widened for the new parameter. Full suite: 515 passed, 12 skipped. Ruff clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,45 @@ uses the test shim and so only exercises the `<sup>` tag route. The resolver
 contract is unit-tested with an injected stylizer, as with the other CSS
 properties.
 
+### Image optimization gates on byte size (#55)
+
+Image optimization only ever triggered on pixel dimensions:
+
+```python
+if max(size) <= max_dim:   # DEFAULT_MAX_DIM = 2048
+    return data
+```
+
+An image can sit well inside that budget and still be far too heavy — encoded
+near-lossless, with EXIF, Photoshop and ICC blocks attached. A 1161x1800 cover
+ran to 2.25 MB and passed through untouched, and because *every* image in that
+book was under 2048px, **not one of its 16 images was optimized — 7.7 MB
+embedded verbatim.**
+
+`optimize_image` now has two independent triggers: dimensions over `max_dim`, or
+encoded weight over `max_bytes` (default 1 MiB, override with
+`KFXGEN_IMAGE_MAX_BYTES`). When only the byte trigger fires the image is
+re-encoded **at its existing dimensions** — it is too heavy, not too big, and
+shrinking it would lose detail for no reason. The existing guard against
+returning something larger than the input still applies.
+
+The default is deliberately generous. Measured across the reference corpus (212
+images in three Amazon-produced KFX files) the largest single image is 466 KB and
+none exceeds 512 KB, so 1 MiB is roughly twice the observed ceiling: it catches
+the pathological tail without recompressing images that are merely large. Set
+`KFXGEN_IMAGE_MAX_BYTES=512000` to track Amazon's own output more tightly.
+
+On the book that surfaced this: 7,677,756 → 4,350,071 bytes (43% smaller, no
+dimension changes), cover 2,254,253 → 490,594.
+
+**Not claimed:** that this fixes the reported "cover does not display on Kindle".
+That symptom is what surfaced the defect, and the sizes are conspicuous, but the
+causal link is unconfirmed — the competing explanation is the unusual
+EXIF/Photoshop/ICC header stack rather than the byte count. Both are resolved by
+re-encoding, so the fix is the same either way. A controlled A/B (identical
+build, single env var) is pending a device sideload to settle it.
+
+
 ## 5.5.0 — Font-embedding toggle (#15) + bold/italic face fix (#50)
 
 **Fixed (#50):** bold and italic embedded faces now render on-device. The applied
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,12 +79,20 @@ the pathological tail without recompressing images that are merely large. Set
 On the book that surfaced this: 7,677,756 → 4,350,071 bytes (43% smaller, no
 dimension changes), cover 2,254,253 → 490,594.
 
-**Not claimed:** that this fixes the reported "cover does not display on Kindle".
-That symptom is what surfaced the defect, and the sizes are conspicuous, but the
-causal link is unconfirmed — the competing explanation is the unusual
-EXIF/Photoshop/ICC header stack rather than the byte count. Both are resolved by
-re-encoding, so the fix is the same either way. A controlled A/B (identical
-build, single env var) is pending a device sideload to settle it.
+**Device-verified.** Sideloading a controlled pair (identical build and book,
+differing only by `KFXGEN_IMAGE_MAX_BYTES`, with distinct titles and ASINs so the
+Kindle could not dedupe them) confirmed it: the 2.25 MB cover does not render, the
+490 KB one does.
+
+The failure turned out to be narrower than expected. The **library thumbnail
+displays for both** — it is only the full-page in-book cover render that fails on
+the heavy image, so that path is evidently on a tighter decode budget than
+whatever generates thumbnails.
+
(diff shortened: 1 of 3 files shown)


jloutsch:
## Device-verified ✅

Sideloaded the controlled pair — identical build and book, differing only by `KFXGEN_IMAGE_MAX_BYTES`, with distinct titles and ASINs so the Kindle could not dedupe them.

| | cover | in-book cover page | library thumbnail |
|---|---|---|---|
| **A** (gate off) | 2,254,253 B | ❌ missing | ✅ shows |
| **B** (1 MiB gate) | 490,594 B | ✅ **shows** | ✅ shows |

Same 1161x1800 dimensions in both. **The fix works.**

## The diagnosis is narrower than the issue assumed

The library thumbnail renders for **both** files. So the same 2.25 MB resource is fine as a thumbnail and fails only as the full-page in-book cover — that render path is evidently on a tighter decode budget than whatever generates thumbnails. Worth recording, because it means "cover missing" is not one failure mode: a missing *thumbnail* would be a different bug entirely and this fix would not address it.

## What is still not isolated

Re-encoding changed **two** variables at once: the byte count (2.25 MB → 490 KB) and the ~20 KB of EXIF + Photoshop APP13 + ICC + Adobe APP14 headers, which `scale_image` strips. Either could be the mechanism.

This does not affect whether to merge — the fix resolves both — but it does affect coverage: **a small image bloated only by header baggage would slip past a byte-size gate.** If that turns out to matter, the follow-up is to strip metadata on re-encode regardless of size, tracked separately.

A lossless header-strip variant (metadata removed, ~2.23 MB, no recompression) would settle it in one more sideload. Happy to build it if the distinction is worth the round trip.

Changelog updated to state the verified result and this limitation rather than the earlier hedge.

Issue #60, opened 2026-07-29
hidden="hidden" navs (page-list, landmarks) are extracted as body content


jloutsch:
## Symptom

A "Page List" section with entries `VII, viii, 1, 2, 3, 4, 5, 6 …` appears in the
book after the table of contents. It is not visible content in the source EPUB.

## Cause

EPUB 3 navigation documents carry non-rendered navs:

```html
<nav class="hidden-tag" epub:type="landmarks" hidden="hidden"> … </nav>
<nav epub:type="page-list" hidden="hidden" role="doc-pagelist">
  <h2>Page List</h2>
  <ol><li id="pageVII"><a href="005_BM_001.xhtml#pageVII">VII</a></li> …
```

`extract_blocks_from_html` walks the DOM without consulting the `hidden`
attribute, so every one of those `<li>`s becomes a content block.

In the affected book that is **374 junk blocks** — `Navigation`,
`Begin Reading`, `Table of Contents`, `Page List`, then several hundred bare
page numbers. 310 of the 328 blocks extracted from `toc.xhtml` come from hidden
navs.

Minimal repro:

```python
extract_blocks_from_html(doc('<p>visible</p><nav hidden="hidden"><p>SHOULD NOT APPEAR</p></nav>'))
# -> ['visible', 'SHOULD NOT APPEAR']
```

## Fix

Skip subtrees carrying the HTML5 `hidden` attribute, and skip navs whose
`epub:type` is a non-rendered kind (`page-list`, `landmarks`). The `hidden`
attribute is the general rule; the `epub:type` check catches producers that omit
it.

## Not a regression

Identical in pre-5.6.0 builds — 374 such blocks in both, verified by extracting
from an older `.kfx`. The in-body link work (#53) merely made it obvious: the
page numbers are now underlined links rather than anonymous stray text.

jloutsch:
Shipped in **v5.6.0** and device-verified: the spurious page list no longer appears after the contents.

`hidden="hidden"` subtrees and navs whose `epub:type` names a non-rendered kind are skipped. On the book that surfaced this, `toc.xhtml` went from 328 extracted blocks to 39, with page-list contamination 310 → 0. Now pinned by the `publisher_structure` golden fixture, verified by reverting the fix and confirming the fixture fails.

Left open after merge because the squash from a stacked branch did not trigger auto-close.

Issue #59, opened 2026-07-29
Tail text after a nested inline tag loses the enclosing element flags (breaks emphasis and links)


jloutsch:
## Symptom

Only part of a linked TOC entry is actually a link. On device: "only the number
of that chapter is underlined showing it is a link" — the chapter title text
beside it is inert.

The same defect silently corrupts **emphasis**, which is the more serious half:
text following a nested tag inside `<em>`/`<strong>` loses its italic/bold.

## Cause

In `converter.py::_walk_inline`:

```python
for child in elem:
    ...
    parts.extend(_walk_inline(child, cur, ...))
    if child.tail:
        parts.append((child.tail, flags))   # <-- flags, not cur
```

`flags` is the *incoming* parameter; `cur` is the accumulated set for this
element (its own italic/bold/super/link plus inherited). A child's tail text
sits inside `elem`, so it must carry `cur`. Using `flags` drops whatever `elem`
itself contributed.

Minimal repro — ` gamma` is inside `<em>` and loses italic entirely:

```python
extract_blocks_from_html(doc("<p><em>alpha <b>beta</b> gamma</em></p>"))
# text  = 'alpha beta gamma'
# spans = [(0,6,italic), (6,4,italic+bold)]     ' gamma' has NO span
```

Real TOC markup hits it through links:

```html
<a href="..."><span class="sans">PART I</span> T<span class="SC">he</span> N<span class="SC">ucleus</span> ...</a>
```

produces link spans over `PART I`, `he`, `ucleus` — but not the tails `T`, ` N`,
` R` between them. Hence a partially underlined, partially dead entry.

## Fix

Pass `cur` for the tail:

```python
parts.append((child.tail, cur))
```

## Impact

Pre-existing and not a regression — it has been mis-rendering emphasis since
inline emphasis landed, silently, because a missing italic is easy to overlook.
The link work (#53) made it visible by rendering the same defect as a broken
link. Worth a regression test for both the emphasis and link cases.

jloutsch:
Shipped in **v5.6.0** — `converter.py:259` now passes `cur` rather than the incoming flags to a child tail.

The visible symptom (half-underlined TOC entries) was device-verified. The more consequential half was silent: text after a nested tag inside `<em>`/`<strong>` had been losing its emphasis since inline emphasis first landed, in every book.

Left open after merge because the squash from a stacked branch did not trigger auto-close.

PR #61, opened 2026-07-29
fix: TOC extraction — nested lists, tail flags, hidden navs (#58, #59, #60)

### LINKED ISSUE #58: Nested <ol>/<ul> inside <li> collapses a whole sub-list into one paragraph
## Symptom

In a book whose EPUB TOC uses nested lists, a Part heading and every chapter
under it render as **one run-on paragraph** instead of separate lines.

Observed on device: "Part 1 and the chapters within it as 1 paragraph".

## Cause

`extract_blocks_from_html` builds `block_tags` from `p div h1..h6 blockquote li
section article figure`. **`ol` and `ul` are not in that set.** The leaf test is:

```python
is_block = elem.tag in block_tags
has_block_child = any(child.tag in block_tags for child in elem)
```

For `<li class="toc-part"><a .../><ol><li>Ch1</li><li>Ch2</li></ol></li>` the
children are `<a>` and `<ol>` — neither is a block tag — so `has_block_child` is
False, the `<li>` is treated as a leaf, and `_walk_inline` flattens the entire
nested list into a single paragraph.

Minimal repro:

```python
extract_blocks_from_html(doc("<ol><li>Part<ol><li>Ch1</li><li>Ch2</li></ol></li></ol>"))
# -> 1 block: 'PartCh1Ch2'      expected: 3 blocks
```

## Fix

Add `ol` and `ul` to `block_tags` so a list counts as a block child and its
parent `<li>` is no longer treated as a leaf.

## Not a regression

Present identically in builds predating 5.6.0 — verified by extracting the same
merged block from a pre-5.6.0 `.kfx`. The in-body link work (#53) only made it
conspicuous, by turning the entries into visible links.

jloutsch:
Fixes #58, #59, #60. Stacked on #54 (they touch the same `_walk_inline`).

Device report: the contents page showed a Part and all its chapters as one run-on paragraph with only the chapter number underlined, plus a stray "Page List" of `VII, viii, 1, 2, 3…` afterwards.

## None of this is a regression

I checked before assuming. Extracting from a **pre-5.6.0** `.kfx` of the same book gives byte-for-byte the same damage — 374 page-number blocks and the identical merged `PART I …` paragraph. The link work in #54 did not cause any of it; it made it visible, by rendering the damage as underlined links instead of anonymous grey text.

## #58 — nested lists flattened

`block_tags` had `li` but not `ol`/`ul`, so `<li>Part<ol><li>Ch1</li></ol></li>` looked childless and the whole sub-list was flattened.

Fixing it exposed the other half of the same bug: a container's own inline text was **dropped** when it also had block children, because that branch only recursed. Inline runs are now flushed as their own blocks in document order — so `<div>lead-in<p>…</p></div>` stops losing "lead-in" too.

## #59 — tail text lost its enclosing flags

```python
parts.append((child.tail, flags))   # incoming flags, not this element's
```

The tail is *inside* `elem`, so it must carry `cur`. This has been silently mangling **emphasis** since inline emphasis landed — in `<em>alpha <b>beta</b> gamma</em>`, ` gamma` was not italic. Nobody noticed because a missing italic is easy to overlook; a half-underlined link is not. That is the more valuable half of this PR.

## #60 — hidden navs rendered

`hidden="hidden"` was ignored, so EPUB 3 `page-list` and `landmarks` navs became body text — one block per printed page.

Skips on the HTML5 `hidden` attribute (general rule) and on `epub:type` naming a structural nav (catches producers that omit `hidden`).

## Result on the real book

| | before | after |
|---|---|---|
| blocks from `toc.xhtml` | 328 | **39** |
| page-list junk blocks | 310 | **0** |
| `PART I` entry | run-on paragraph, digit linked | own block, **single span over all 38 chars** |

## Tests

7 new, written failing first, each pinned to a minimal repro. Full suite 567 passed, 12 skipped. Ruff clean. Golden corpus unchanged.

**Not yet device-verified** — needs a sideload.

### DIFF
--- a/plugin/kfxgen/converter.py
+++ b/plugin/kfxgen/converter.py
@@ -220,7 +220,13 @@ def _walk_inline(
                 )
             )
         if child.tail:
-            parts.append((child.tail, flags))
+            # The tail sits inside `elem`, so it carries `cur` — this element's
+            # accumulated flags — not the incoming `flags`. Using `flags` here
+            # dropped whatever `elem` itself contributed: text after a nested
+            # tag inside <em> lost its italic, and text between styled spans
+            # inside an <a> lost the link, leaving entries half-underlined and
+            # half-dead. (#59)
+            parts.append((child.tail, cur))
     return parts
 
 
@@ -253,6 +259,26 @@ def _dedupe_keep_order(items):
             seen.add(it)
             out.append(it)
     return out
+
+
+#: epub:type values naming navigation that is never rendered as body text.
+#: page-list is the big one: an EPUB 3 print-pagination nav holds one entry per
+#: printed page, which arrived as several hundred bare page numbers. (#60)
+_NON_RENDERED_EPUB_TYPES = {"page-list", "landmarks", "lot", "loi", "lov"}
+_EPUB_TYPE_ATTR = "{http://www.idpf.org/2007/ops}type"
+
+
+def _is_non_rendered(elem):
+    """True when `elem` heads a subtree that is markup, not reading content.
+
+    Two signals: the HTML5 `hidden` attribute (the general rule — the producer
+    said not to display this), and an `epub:type` naming a navigation kind that
+    is structural by definition, which catches producers that omit `hidden`.
+    """
+    if elem.get("hidden") is not None:
+        return True
+    raw = elem.get(_EPUB_TYPE_ATTR) or elem.get("epub:type") or ""
+    return any(part in _NON_RENDERED_EPUB_TYPES for part in raw.lower().split())
 
 
 def _attach_anchor_keys(blocks, base_href):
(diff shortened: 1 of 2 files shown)


jloutsch:
## Device-verified ✅

Confirmed on hardware as part of the full-stack build: the contents page reads as separate lines (no run-on Part+chapters paragraph), entries link across their whole title rather than just the leading numeral, and the spurious page list is gone.

The #59 half of this PR — tail text after a nested inline tag losing its enclosing flags — is the broadest fix here. It was silently dropping italic in `<em>alpha <b>beta</b> gamma</em>` for every book, not just mangling TOC links.

Corpus evidence from the Gutenberg-90 sweep (see #63) backs this up: the container-inline-text half recovered **44% of the body text** of one book that 5.5.0 was silently discarding, with zero duplication and zero blocks lost.

PR #63, opened 2026-07-29
fix: stop losing anchors for back-matter and whole-file link targets (#62)

### LINKED ISSUE #62: TOC links to back-matter and whole-file targets are silently dropped
Reported from device: several TOC entries that are links in the EPUB are inert in the KFX — Afterword, Bibliography, Notes, Also by, About the Author, Discover More.

Extraction is correct in every case; the anchors are lost later, so `$179` has nothing to resolve and the link is dropped by design. Three distinct paths:

**1. Elided title-duplicate blocks take their ids with them.** Back-matter commonly opens with a heading equal to the chapter title (`AFTERWORD` in a chapter titled `AFTERWORD`). That block is dropped as redundant — but it carries the `id` the TOC links to. Its `anchor_keys` must be carried onto the chapter's first chunk.

**2. Image blocks never carried anchor_keys.** Figure ids (`#fig3`) live on image blocks, and the image branch appended the chunk without them, so links to figures resolved to nothing.

**3. Whole-file targets with no fragment.** `<a href="038_BM_006.xhtml">About the Author</a>` — the file-only fallback could only resolve a file that declares at least one id somewhere, and these files declare none. The first block of each document now also carries a bare-filename key.

Measured on the affected book: 20 anchor keys were being lost between the chapter list and the chunk list.

## Note on diagnosis

I initially tested the elision hypothesis and wrongly concluded it was not the cause — the probe called `extract_chapters_from_oeb` without `metadata=`, which yields generic `Section 35` titles that never match the heading. With metadata the real titles match and the elision fires. Worth recording because the wrong conclusion looked well-supported.

jloutsch:
Fixes #62. Stacked on #61.

Six TOC entries were inert on device. Extraction was correct in all six — the anchors were lost between the chapter list and the chunk list. **20 anchor keys** vanished this way on the affected book.

| path | cause |
|---|---|
| elided title-duplicate block | back-matter opens with a heading equal to its chapter title; the block is dropped as redundant, taking the TOC target id with it |
| image blocks | appended without `anchor_keys`, so figure ids (`#fig3`) anchored nothing |
| whole-file targets | `<a href="about.xhtml">` only resolved if that file declared some id; files declaring none were unreachable |

## Result on the real book

All six now link with full character coverage:

```
LINK cov= 9/9   AFTERWORD
LINK cov=13/13  Discover More
LINK cov=12/12  Bibliography
LINK cov=16/16  About the Author
LINK cov=19/19  Also by <author>
LINK cov= 5/5   Notes
```

## Diagnosis note

I tested the elision hypothesis early and **wrongly ruled it out** — my probe called `extract_chapters_from_oeb` without `metadata=`, which yields generic `Section 35` titles that never match the heading. With metadata the real titles match and the elision fires. I had already told the user it was not the cause. Recording it because the wrong conclusion looked well-supported, and only instrumenting the real generator settled it.

## Generality

Smoke-tested across four other EPUBs (English trade non-fiction, Russian, Korean, Japanese): all convert, **zero dangling links**, no nav junk. 571 tests pass, ruff clean, golden corpus unchanged.

### DIFF
--- a/plugin/kfxgen/converter.py
+++ b/plugin/kfxgen/converter.py
@@ -292,6 +292,13 @@ def _attach_anchor_keys(blocks, base_href):
             if normalized
             else []
         )
+    # A TOC entry may link to a whole file with no fragment
+    # (`<a href="about.xhtml">About the Author</a>`). Give the document's first
+    # block a bare-filename key so such links have something to resolve to —
+    # otherwise a file that declares no ids anywhere is unreachable and the
+    # link is silently dropped. (#62)
+    if normalized and blocks:
+        blocks[0]["anchor_keys"] = [normalized] + blocks[0]["anchor_keys"]
     return blocks
 
 
--- a/plugin/kfxgen/native_generator.py
+++ b/plugin/kfxgen/native_generator.py
@@ -40,6 +40,17 @@ SUPERSCRIPT_SHIFT = ("35", "$314")  # +35% of font size
 SUBSCRIPT_SHIFT = ("-20", "$314")  # -20% of font size
 
 
+def _dedupe_keys(keys):
+    """Order-preserving dedupe for anchor-key lists."""
+    seen = set()
+    out = []
+    for k in keys:
+        if k not in seen:
+            seen.add(k)
+            out.append(k)
+    return out
+
+
 def _safe_write_bytes(path, data):
     """Write `data` to `path` with symlink and traversal defenses (#45).
 
@@ -2471,6 +2482,10 @@ class NativeKFXGenerator:
 
         for ch_idx, chapter in enumerate(chapters):
             start_idx = len(all_chunks)
+            # Anchor ids belonging to blocks this chapter drops (a heading that
+            # duplicates the chapter title). Re-attached to the chapter's first
+            # chunk below so links to those ids still resolve. (#62)
+            carried_anchor_keys = []
             title = chapter["title"]
             toc_links = chapter.get("toc_links")
 
@@ -2544,6 +2559,15 @@ class NativeKFXGenerator:
                                         "spans": rebased_spans,
                                     }
                                 else:
+                                    # The block is gone, but its ids are what
+                                    # the TOC links to — a back-matter file
+                                    # opening with a heading equal to its
+                                    # chapter title is the common shape. Carry
+                                    # them onto the chapter's first chunk so
+                                    # the link still lands here. (#62)
+                                    carried_anchor_keys.extend(
+                                        first.get("anchor_keys") or []
+                                    )
                                     iter_blocks = iter_blocks[1:]
                         para_iter = iter_blocks
                     else:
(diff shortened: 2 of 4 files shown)


jloutsch:
## Gutenberg top-90 corpus sweep

Ran all 90 books through **both** a 5.5.0 baseline build and this stack (#54 → #61 → #63), in separate isolated Calibre configs. 180 conversions, **0 failures**.

| | baseline 5.5.0 | this stack |
|---|---|---|
| conversion failures | 0 | 0 |
| `$266` anchors carrying `$180` | **0** | 64,820 |
| `$179` link spans | 48,498 | 74,606 |
| dangling link targets | 48,498 (all of them) | **0** |
| nav junk blocks | 0 | 0 |
| books losing text | — | **0** |

### Every link in the corpus was dead before

The baseline emits 48,498 link spans and **zero** anchors carrying `$180`. That is #51 reproduced across 90 unrelated books: every internal link in every book kfxgen has ever produced resolved against nothing.

### Recovered text — 7 books

No book lost text. Seven gained it:

```
85_The_Lady_of_the_Lake      257,784 -> 462,714  (+79.5%)
71_Southern_Literature       660,623 -> 811,375  (+22.8%)
54_Color_Images_Mars_Rovers   44,432 ->  62,231  (+40.1%)
73_Beowulf                   229,627 -> 250,803   (+9.2%)
65_Carmen                    140,537 -> 146,460   (+4.2%)
43_Works_of_George_Gillespie 1,625,122 -> 1,691,826 (+4.1%)
40_History_of_Tom_Jones      1,921,238 -> 1,966,153 (+2.3%)
```

I checked this was real recovery rather than duplication before reporting it. On *The Lady of the Lake*: **0** blocks duplicated, **0** baseline blocks missing, and the 391 new blocks are the poem's stanzas — `Harp of the North! that mouldering long hast hung…`, `I. The stag at eve had drunk his fill…`. The baseline was dropping 44% of that book's body text.

That is the container-inline-text half of #61: a block element's own text was discarded whenever it also had block children. Verse and nested-markup books hit it hardest.

### Caveat

Gutenberg EPUBs are structurally simpler and more uniform than commercial publisher output — no DRM-era pagination navs, lighter CSS. The sweep is strong evidence against regressions and for the recovery fixes, but it is not a substitute for testing publisher EPUBs, which is where #58/#60 originally came from.

jloutsch:
## Device-verified ✅

Sideloaded a build carrying the full stack (#54 → #61 → #63 plus #57). All six previously-inert TOC entries — Afterword, Discover More, Bibliography, About the Author, Also by, Notes — now link correctly on hardware.

Together with the earlier confirmations this leaves the whole stack verified on-device:

| PR | fixes | status |
|---|---|---|
| #54 | #51 anchors, #52 superscript, #53 in-body links | ✅ device |
| #61 | #58 nested lists, #59 tail flags, #60 hidden navs | ✅ device |
| #63 | #62 anchor loss (elided headings, image blocks, whole-file targets) | ✅ device |
| #57 | #55 image byte gate | ✅ device |

The two entries that remain unlinked are correct: they target an external publisher URL, external URLs with no in-book anchor to resolve against.

Still unverified anywhere: **subscript** (`$31` negative shift). No reference file in the corpus uses one and no test book exercises it.

PR #81, opened 2026-08-01
fix: split $145 content fragments to stay under the format maximum (#37)

### LINKED ISSUE #37: Content () fragments exceed the 8192-byte maximum for long chapters
## Summary

Decoding kfxgen output with jhowell's `kfxlib` reports many content fragments
over the format's per-fragment maximum:

```
ERROR: Content content_104: 11790 bytes exceeds maximum (8192 bytes)
ERROR: Content content_7:  121371 bytes exceeds maximum (8192 bytes)
...
```

Seen across three ordinary-size trade books (an English title, a CJK reference
title, and a CJK novel that embeds fonts) — not just pathological
high-chapter-count books. Individual `$145` fragments reach ~121 KB, ~15× the
8192-byte limit.

## Cause (suspected)

`generate_full_book` builds one `$145` content fragment per chapter from that
chapter's chunk-text slice, without splitting the aggregate to stay under the
per-fragment byte cap. Chunking (`CHUNK_SIZE`) governs `$259`/position
granularity, but the `$145` payload for a long chapter is emitted whole. Real
KDP/`KFX Output` splits content across multiple fragments per section.

## Impact (needs device confirmation)

Unknown whether the Kindle firmware tolerates oversized `$145` fragments or
truncates/fails them. `PROGRESS.md` historically treats "content exceeds
maximum" as benign decoder scale-noise, but that was for huge books; here it's
single fragments an order of magnitude over the cap on normal books. **Risk: long
chapters render truncated on-device.** Should verify by sideloading a book with a
long chapter and checking the chapter renders in full.

## Fix direction

Split a chapter's `$145` content into multiple fragments (each <= 8192 bytes) at
chunk boundaries, updating the `$259`/`$260`/position references and the
`$270`/`$419` entity maps accordingly, matching the reference multi-fragment
layout. Add a decode-based assertion (kfxlib available via the installed
`KFX Input.zip`) that no emitted `$145` exceeds the cap.

## Context

- Independent of #15 (font embedding); surfaced during non-Roman conversion
  review. Related to the position/section-map work in #20 / #30.

jloutsch:
Closes #37.

## Reproduced first

Against upstream `kfxlib`, not by reasoning:

```
Content content_1: 25240 bytes exceeds maximum (8192 bytes)
```

## The metric is not what the design assumed

This is the part that would have shipped a wrong fix. From upstream `kfxlib`'s `yj_structure.py`:

```python
for content in fragment.value["$146"][:-1]:
    content_bytes += len(content.encode("utf8"))
if content_bytes >= MAX_CONTENT_FRAGMENT_SIZE:
```

The **last string is excluded** from the total, and the comparison is **`>=`** — a fragment measuring exactly 8192 is already a violation, not the last passing value. #37's design said "each <= 8192" and "boundary at exactly 8192", both of which are wrong against what upstream enforces. A splitter budgeting the obvious way passes its own tests and still fails upstream.

One consequence works in our favor: a group holding a single string is always conformant however long that string is, so there is no unsplittable input and the packer always makes progress.

## Changes

- **`pack_content_fragments()`** groups a chapter's chunks under the cap, budgeting on encoded bytes. CJK is 3 bytes/char — a character budget clears the ASCII chapter and still overflows.
- **`build_fragment_259`** takes a per-entry content reference instead of one fragment name plus a running index. That single-name assumption was what blocked splitting. `content_index_offset` was removed with it; it was always 0.
- Names are allocated **sequentially across the book**, so a book whose chapters each fit still produces exactly `content_1..content_N`.
- All split fragments are registered in `$419` and `$270` (verified: 7 fragments, 7 entity-map entries).

## Verification

| Check | Result |
|---|---|
| Six pre-existing goldens | **byte-identical** — books under the cap are untouched |
| Position map, split vs unsplit | `$265`, `$264`, `$550`, `$260` and all 182 `$155` ids **identical**; only `$259` differs |
| Upstream kfxlib, all 7 fixtures | **0** "exceeds maximum" messages |
| Disabling the split | **4 tests fail** across tier2 and tier3 |
| Suite | 635 passed; 69 under tier2/tier3/tier3_strict |
| Ruff | `check` and `format --check` clean |

#37 flagged the position map as the open risk and said it "must be verified, not assumed". It's verified: I built `long_chapter` with splitting on and off and diffed the fragments. Splitting changes only content references.

New `long_chapter` golden covers the split path with both an ASCII and a CJK chapter, plus a shape test asserting it stays several fragments' worth so it can't rot into a short book.

## Not verified

**Device.** Your July comment established the oversized fragments are tolerated by current firmware — all three books rendered correctly. So this turns working output into conforming output, and the change touches the content reference on every text entry in every book. Structural evidence is as strong as I can make it locally, but per the project's own rule this is a hypothesis until a long-chapter book is sideloaded and confirmed to render in full with navigation intact. I'd suggest not merging until then.


jloutsch:
## Device verification (physical Kindle, sideloaded)

The gap this PR was held open for. Test file: a commercial trade title, 918,910 characters, 27 chapters — converted through `ebook-convert` with a plugin built from this branch.

Baseline for the same book on `main`: **19 of 28** content fragments over the cap, largest 99,063 bytes (12x). On this branch: **128 fragments, 0 over cap**, largest 8,190.

### Boundary reads

19 chapters now span 6-13 fragments. Read *through* the split points rather than spot-checked, since a mis-mapped index would serve a plausible-looking wrong paragraph rather than crash:

| Location | Result |
|---|---|
| Ch. 1, `content_12` -> `content_13` | reads clean |
| Ch. 10, ¶13 -> ¶14 (`content_66` -> `content_67`) | reads clean |
| Ch. 10, `content_67` -> `content_68` | reads clean |

### Link round-trips

Both into the endnotes, which went from a single 99 KB fragment to 13 — the densest link target in the book:

- Superscript **1** (Hemingway epigraph) -> landed on `1 'Notes on the next war', Esquire, September 1, 1935`. Correct.
- Superscript **2** (General Bydén) -> landed on the page holding note 2; back-navigation returned to the correct place in ch. 1. Correct.

### Structural, same book, `main` vs branch

- Reading order **identical** — 3,944 paragraphs, 910,818 characters (walked via `$259`, not file order)
- Positions **identical** — all 3,948 `$155` entries
- Links **identical** — 4,618 spans, 2,367 anchors, 0 dangling; all 4,618 resolve to a real paragraph
- Only fragment-type count that moved: `$145`, 28 -> 128
- Upstream kfxlib: **0** "exceeds maximum", 0 error-level messages

### One thing worth recording

A raw file-order text comparison reports the two builds as different. That is fragment *ordering*, not content: the container sorts `$145` fragments lexicographically, so `content_1, content_10, content_100, ...` only diverges from numeric order once a book has more than 9 — which never happened before this change. `$259` addresses fragments by name, so reading order is unaffected. Anyone diffing two builds by concatenating `$145` in file order will see a false positive.

Device gate met.

Issue #92, opened 2026-08-02
Kindle Previewer does not complete under launchd, so the upstream drift check is interactive-only


jloutsch:
Found immediately after #88 shipped, by running the installed agent rather than the script by hand.

## Measurement

Same machine, same script, same Previewer (3.106.0):

| Context | Elapsed | Result |
|---|---|---|
| run by hand | ~9-16s | `current` / `available` — correct |
| run by `launchd` | **100s** | hits the script's 90s ceiling → `unknown`, every time |

Kindle Previewer does not finish when a background agent launches it. Nothing appears on stderr or in the agent's stdout capture; the process simply sits until killed. The first time this happened it ran unbounded for minutes before I killed it by hand, which is what prompted the timeout in #90.

Most likely a GUI-session constraint on driving an Aqua application from a `launchd` agent, but that is a hypothesis — it has not been confirmed.

## Current state

The shipped plist sets `KFXGEN_DRIFT_SKIP_UPSTREAM=1`, so the scheduled job runs the offline half only and completes in ~2s instead of burning 90s to report `unknown`. The log records `upstream=skipped`, which is honest about what did and did not happen.

This is the right default: a check that can only ever return `unknown` is worse than no check, because it looks installed and healthy while reporting nothing — the failure shape this repo keeps finding (#76, #78, #79). Better to disable it visibly than to leave it running and useless.

The offline comparison is unaffected and remains the half that has actually caught drift (3.98.0 → 3.106.0).

## Worth trying

Untested, roughly in order of effort:

1. `LimitLoadToSessionType: Aqua` in the plist — the default for a LaunchAgent should already give a GUI session, so this may be a no-op, but it is cheap to rule out.
2. `launchctl asuser $UID` wrapping the Previewer invocation.
3. Check whether Previewer needs a window server connection even for `-log`, which would make it fundamentally undrivable from a background agent. If so, close this as won't-fix and keep the upstream check documented as interactive.
4. A different upstream signal entirely that does not require running Previewer — though #88 established there is no cached update state on disk, and `-update` installs rather than reports.

## Priority

Low. The scheduled job works and does the useful part; this is about recovering a secondary signal.

jloutsch:
## Option 1 ruled out: `LimitLoadToSessionType: Aqua` does not help

Time-boxed experiment, run against the live agent rather than in isolation:

1. Added `LimitLoadToSessionType = Aqua` to the installed plist
[...]

Issue #91, opened 2026-08-02
Previewer 3.106 emits YJ_symbols max_id 844; installed kfxlib knows 843


jloutsch:
Observed while re-baselining for the 3.98.0 → 3.106.0 drift (#88). Every decode of Previewer 3.106.0 output warns:

```
Import symbol table YJ_symbols version 10 max_id 844(+9=853) exceeds known table size 843(+9=852)
Unknown symbols: max_id=853
```

Amazon extended the shared `YJ_symbols` table by one entry. The installed `kfxlib` (20260520, KFX Input 2.33.0) predates it, so it decodes the file but cannot name the new symbol.

## Why this is separate from the drift itself

The drift diff found one new symbol, `$23`, in 3.106.0's prose output. That is a symbol Amazon *used*. This issue is about the symbol *table* growing — a different fact, and the one that suggests a newer KFX Input plugin may exist.

## Is anything broken?

No, and the reasons are worth stating rather than assuming:

- kfxgen writes its own fragments and never emits `$23`
- the vendored `kfxlib_minimal` decodes kfxgen's own output, which does not contain the new symbol
- tier-2 differential decode passes (43 tests) against 20260520
- the warning is emitted by upstream `kfxlib` decoding **Amazon's** output, not ours

So this is a watch item, not a defect.

## What to check

1. Whether a KFX Input newer than 2.33.0 / kfxlib 20260520 exists that knows symbol 844. If so, refresh `tests/fixtures/vendor/kfx_input_plugin.zip` and its version pin, and re-run tier-2.
2. Whether the new symbol matters to `kfxlib_minimal`'s audit table. Since kfxgen neither emits nor needs to parse it, probably not — but the audit table should reflect the current upstream surface rather than silently lag it.

## Priority

Low. Filed so the observation is not lost between the drift baselines and the vendored pin, which are maintained in different places and would each individually look fine.

jloutsch:
## Checked the GitHub mirror — it is behind, not ahead

`kluyg/calibre-kfx-input` describes itself as *"a GitHub mirror of John Howell's (jhowell) KFX Input Calibre plugin"* (GPL-3.0). It looked like a way to answer this issue's open question. It is not:

| | Version | kfxlib |
|---|---|---|
| mirror, latest commit `2025-07-29` | 2.25.0 | `20250519` |
| installed locally | **2.33.0** | **`20260520`** |

[...]

PR #98, opened 2026-08-07
feat: watch the calibre plugin index for a newer KFX Input (#91)

jloutsch:
Addresses the first action item in #91.

## Why

The vendored kfxlib pin and the drift baselines are maintained in different places, so each looks current on its own while the plugin that produced them has moved. Noticing a new KFX Input was a manual step nobody was scheduled to take.

The version ladder as measured today:

| | YJ symbols |
|---|---|
| `kfxlib_minimal` (ours) | 842 |
| upstream kfxlib 20260520 | 843 |
| Previewer 3.106.0 output | 844 |

## What it does

Compares the installed plugin against calibre's own plugin index — the same JSON its updater reads — and reports `current`, `available`, `unknown`, or `skipped`. It never installs.

**This one runs on the schedule**, unlike the Previewer probe disabled in #92. That probe fails under `launchd` because Kindle Previewer will not finish when a background agent launches it; fetching a file has no GUI dependency. That is the entire difference, and it is now written down next to both checks so the distinction survives.

Knobs: `KFXGEN_DRIFT_SKIP_PLUGIN_INDEX=1` disables it; `KFXGEN_PLUGIN_INDEX_URL` overrides the index.

## The bug worth reading

The first version produced a **false all-clear** — the "a newer plugin exists" case reported `current` — and only in the environment the schedule actually runs in.

`run_bounded`'s fallback watchdog backgrounds its command, and a background job in a non-interactive shell has stdin redirected from `/dev/null`. The Python was fed by heredoc, so it arrived empty, exited 0 having done nothing, and the check reported success. The fallback is only reached where `timeout` is absent — which is stock macOS. By hand, with Homebrew coreutils on `PATH`, it worked fine.

Fixed by passing the script with `python3 -c`. Documented in both the script and the README, since this is the same failure shape as #92: a check that looks installed and healthy while reporting nothing.

## Verification

All four states, in both environments — `timeout` present, and a bare `PATH` with `env -i`:

| index says | with `timeout` | minimal env |
|---|---|---|
| 2.33.0 (same) | exit 0, `current` | exit 0, `current` |
| 2.34.0 (newer) | exit 1, `available` | exit 1, `available` |
| unreachable | exit 0, `unknown` | exit 0, `unknown` |
| skip flag set | `skipped` | `skipped` |

A lower version in the index does not fire. Driving these states needs no release — `KFXGEN_PLUGIN_INDEX_URL` takes a local `file://` bz2.

As of today the index reports KFX Input **2.33.0**, matching both the committed pin and what is installed, so the check is quiet.

## Not included

The other findings from reviewing #91 — the 842/843 correction, and that `kfxgen` does emit `$23` contrary to the issue text — are posted as a comment there rather than changed here, since neither requires a code change.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -75,18 +75,19 @@ def test_nav_pane_renders_on_paperwhite():
 Composite categorization markers (`unit`, `integration`, `slow`, `benchmark`,
 `critical`) are orthogonal to tier — use them in addition, not instead.
 
-## Vendored kfxlib
+## The upstream kfxlib copy
 
 The tier-2 differential decode test (`tests/integration/test_kfxlib_diff.py`)
-compares output from our vendored `kfxlib_minimal` subset against Calibre's
+compares output from our trimmed `kfxlib_minimal` copy against Calibre's
 upstream `kfxlib` (jhowell's KFX Input plugin). That third-party plugin is
 **not redistributed in this repository**, and the reason is narrower than "the
 license forbids it": the `kfxlib` *source* inside it is marked GPL v3, which is
-the grant kfxgen already relies on to vendor a modified subset (see `NOTICE`).
-What is not redistributed is the *packaged plugin zip* — a third-party build
-artifact that ships with no `LICENSE` file and no package-level terms, so
-nothing states how it may be passed on. To run the tier-2 test, supply the zip
-locally at
+the grant kfxgen already relies on to include a modified copy of part of it
+(see `NOTICE`). What is not redistributed is the *packaged plugin zip* — a
+third-party build artifact that ships with no `LICENSE` file and no
+package-level terms, so nothing states how it may be passed on.
+
+To run the tier-2 test, supply the zip locally at
 `tests/fixtures/vendor/kfx_input_plugin.zip` using the procedure below; the
 test skips cleanly when the file is absent, so CI without it still passes.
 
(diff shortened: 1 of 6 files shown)


jloutsch:
## Verified under the real agent — and it failed twice first

The PR claimed "this one runs on the schedule." Running the installed `launchd` agent **before** merging showed it did not: `index=unknown` on every scheduled run, while working perfectly by hand. That is precisely the #92 failure shape, and it would have landed on `main` as a false claim.

### Root cause: the host, not launchd

`code.calibre-ebook.com` serves an **incomplete certificate chain** — no intermediate — so anything trusting only the OS store fails on it:

```
api.github.com                      http=200
example.com                         http=200
code.calibre-ebook.com/plugins/...  curl: (60) unable to get local issuer certificate
```

Two implementations died on this before the cause was clear:

1. `urllib` — blamed on Xcode's `/usr/bin/python3` having no CA bundle, which is true but wasn't the whole story
2. `curl` — same error, which is what ruled out the "python has no certs" theory

### The fix

calibre ships its own CA bundle; that is why its plugin updater can read an index `curl` cannot. The check now calls `calibre.utils.https.get_https_resource_securely` through `calibre-debug`, reading the index exactly as calibre does. calibre is guaranteed present wherever this script has anything to check.

### Now verified

Under the installed agent:

```
ok previewer=3.106 kfxlib=20260520 kfxinput=2.33.0
   (2 baseline(s) current, upstream=skipped, index=current)
```

Remaining states, by hand and in a bare-`PATH` environment:

| | result |
|---|---|
| index newer | exit 1, refresh instructions |
| index same / older | exit 0, `current` |
| unreachable | exit 0, `unknown` |
| calibre missing | exit 0, `unknown` |
| skip flag | `skipped` |

A `file://` URL bypasses the fetch, so none of this waits on a release.

The README now records both traps — the stdin one and the certificate one — and says plainly that a hand-run proves nothing about the schedule.

Issue #99, opened 2026-08-08
Tier-2 differential decode cannot gate anything, and a green CI does not say so


jloutsch:
> **Next step: talk to jhowell.** Decided 2026-08-07. Before building any of
> the workarounds below, ask upstream directly — the code is already GPL v3 and
> the ask is narrow. A yes turns tier-2 into an ordinary CI gate and makes most
> of the options here unnecessary; a no costs one forum post and leaves things
> exactly as they are. Draft message at the bottom of this issue.

Noticed while reviewing #91, which cited "tier-2 passes (43 tests)" as evidence nothing was broken. That was true as a local run, but tier-2 is not a standing gate and never has been.

## The state of it

- `.gitignore:72` excludes `tests/fixtures/vendor/*.zip`. Only `kfx_input_plugin.version.txt` is committed.
- With the zip absent, `tests/integration/test_kfxlib_diff.py` reports **32 skipped**.
- CI never supplies the zip, so those 32 skip on every run, on every PR, and have since the suite was added.

This is deliberate and documented — `CONTRIBUTING.md` → *Vendored kfxlib* says the plugin is "**not redistributed in this repository** — its license does not grant redistribution rights," and that "the test skips cleanly when the file is absent, so CI without it still passes."

So the constraint is real and committing the zip is not on the table. This issue is not "why isn't it committed."

## The actual problem

A green CI reads as "everything that can be checked was checked." For tier-2 that is false, and nothing in the output says otherwise — 32 skips scroll past in a log nobody reads line by line. The independent-provenance oracle described at the top of `test_kfxlib_diff.py` is the one designed to catch generation bugs that one decoder accepts and the other rejects, and it contributes nothing to any automated decision.

That is the same failure shape this repo keeps paying for: a check that looks installed and healthy while reporting nothing. #92 disabled a Previewer probe for exactly this reason, and #98 hit it twice more — a heredoc-fed script that exited 0 having done nothing, and a fetch that returned `unknown` on every scheduled run while working by hand.

The risk is not that tier-2 is unavailable. It is that its absence is invisible.

## Options, none obviously right

1. **Make the skip loud.** CI prints a prominent "tier-2 DID NOT RUN — differential decode unverified" line, or reports a neutral (not green) status check. Cheapest, changes no coverage, removes the false impression.
[...]

PR #106, opened 2026-08-21
test: assert the symbol-catalog invariant that actually matters (#91)

### LINKED ISSUE #91: Previewer 3.106 emits YJ_symbols max_id 844; installed kfxlib knows 843

jloutsch:
Closes #91.

## What #91 asked, and the answers

**1. Does a KFX Input newer than 2.33.0 / `kfxlib` 20260520 exist that knows symbol 844?**

No. Answered in the issue thread: 2.33.0 is the newest release on jhowell's MobileRead thread, so upstream is the one lagging Amazon, not us. The GitHub mirror that looked like a way to check is ~a year stale (2.25.0 / `20250519`). #98 added the plugin-index drift watch so the next release is noticed on a schedule rather than by hand.

**2. Does the new symbol matter to `kfxlib_minimal`'s audit table?**

Yes, but not in the direction the issue assumed. Re-measuring against the pinned zip shows the audit table was recording the wrong quantity.

## The measurement

| | symbols | ids | declared `max_id` |
|---|---|---|---|
| `kfxlib_minimal` (ours) | 842 | `$10`–`$851` | 842 |
| `kfxlib` 20260520 (pinned upstream) | 843 | `$10`–`$852` | 843 |
| Kindle Previewer 3.106 output | 844 | — | 844 (`+9=853`) |

The README recorded the drift as *"13 catalog symbols behind (`$835`–`$846`, `$851`)"*. Those 13 entries differ from upstream's **only by a trailing `?`** — an annotation meaning "this id exists but has never been observed in real content". `ion_symbol_table.py:266` strips it on load, so `$835?` and `$835` are the same symbol at the same id. No effect on encoding or decoding. The recorded drift was an artifact of diffing source strings.

The one real difference is table length, and it is one entry, which is itself an unnamed placeholder.

## Why the length is deliberately left alone

`StandardSymbolTable` imports the shared table with `max_id = len(YJ_SYMBOLS.symbols)` — 842. Every reader truncates its own copy to that length (`ion_symbol_table.py:203`), so ids kfxgen emits resolve identically on a device whose table is longer. Confirmed locally: the emitted import is `YJ_symbols version 10 max_id 842`.

Bumping to 843 would shift every local symbol id by one (locals start after the imported range) and rewrite every generated file, in exchange for a placeholder kfxgen never emits. Against a format whose only real verification is a device sideload, not worth it.

## What this PR adds

The assumption holding all of the above together — *upstream only ever appends; our prefix is identical* — was untested. If Amazon renamed or renumbered an id inside the range we already declare, every `$NNN` kfxgen emits would silently mean something else, and nothing would catch it.

`test_yj_symbol_catalog_is_prefix_of_upstream` (tier-2) asserts the prefix property: same table name and version, our length `<=` upstream's, identical names for every id we declare.

Asserting a prefix rather than a symbol count is the point. Amazon appends to this table as firmware evolves, so a count assertion would fail on every upstream release while telling us nothing. This one stays quiet through appends and fires only on the change that would corrupt output.

## Verification

- Mutation-tested: renaming one entry in our catalog fails with `First offenders: [(77, 'renamed_by_amazon', '$77')]`; revert passes.
- 46 tier-2 tests pass with the vendored zip present (45 existing + 1 new).
- 581 unit tests pass.
- `ruff check` and `ruff format --check` clean (0.15.1).

Note the standing #99 caveat: tier-2 skips in CI, so this test — like the other 45 — only runs where the KFX Input zip is present.

## Not changed

`docs/kfx-generation-explained.md` already states 842/843/844 correctly. No behavior change, so no version bump or CHANGELOG entry, matching the pattern of other docs/test-only commits.

### DIFF
--- a/plugin/kfxgen/kfxlib_minimal/README.md
+++ b/plugin/kfxgen/kfxlib_minimal/README.md
@@ -13,8 +13,37 @@ copyright "2016-2025, John Howell"). Update this on every re-sync.
 | Field | Value |
 |-------|-------|
 | Fork baseline (approx.) | jhowell `kfxlib` ≈ late-2025; the fork carries no `version.py` |
-| Latest upstream compared | `kfxlib` **20260520** (KFX Input 2.33.0) — measured 2026-07-04 |
-| Drift at that comparison | 13 catalog symbols behind (`$835`–`$846`, `$851`); `YJ_symbols` table version **10** (unchanged); `ROOT_FRAGMENT_TYPES` identical. **Benign for generation** — kfxgen emits none of the missing symbols. |
+| Latest upstream compared | `kfxlib` **20260520** (KFX Input 2.33.0) — re-measured 2026-08-21 (#91) |
+| Drift at that comparison | `YJ_symbols` table version **10** (unchanged); `ROOT_FRAGMENT_TYPES` identical. Our catalog holds **842** symbols (`$10`–`$851`), upstream **843** (`$10`–`$852`). The one extra entry is itself an unnamed placeholder. **Benign for generation.** |
+
+### `YJ_symbols` catalog: what the numbers mean (#91)
+
+The earlier reading of this drift as "13 catalog symbols behind (`$835`–`$846`,
+`$851`)" was measuring the wrong thing. Those 13 entries differ from upstream's
+only by a trailing `?`, which is an annotation meaning "this id exists but has
+never been observed in real content" — `ion_symbol_table.py:266` strips it on
+load, so `$835?` and `$835` are the same symbol at the same id. The annotation
+has no effect on encoding or decoding. The only real difference is table length.
+
+Length is safe to lag, and deliberately not bumped. `StandardSymbolTable`
+imports the shared table with `max_id = len(YJ_SYMBOLS.symbols)` — currently
+**842** — and every reader truncates its own copy to that length, so the ids
+kfxgen emits resolve identically on a device whose table is longer. Raising it
+to 843 would shift every local symbol id by one and rewrite every generated
+file, in exchange for a placeholder kfxgen never emits. Against a format whose
+only real verification is a device sideload, that is a bad trade.
+
+For scale: Kindle Previewer 3.106 declares `max_id` **844** (`+9=853`), which is
+why upstream `kfxlib` 20260520 warns when decoding Amazon's own output. That
+warning is about upstream's table, not ours, and appears on Amazon-produced
+files only.
+
+What *would* break generation is upstream renaming or renumbering an id inside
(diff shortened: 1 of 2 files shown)


jloutsch:
Added a second test after re-reading #91's closing note — *"maintained in different places and would each individually look fine."*

`tests/fixtures/vendor/kfx_input_plugin.zip` is not committed, so `kfx_input_plugin.version.txt` is the only committed record of which upstream this repo was checked against. The audit table in `kfxlib_minimal/README.md` and the drift watch in `research/kfx-format-baseline/` both reason from that sidecar, and nothing tied it to the zip. Refresh one without the other and both files still look plausible while every conclusion drawn from them is wrong.

`test_vendored_pin_matches_the_zip_it_describes` reads `kfxlib/version.py` out of the zip and asserts the sidecar matches. Skips with the zip absent, same as the rest of tier-2.

Mutation-verified: setting the pin to `20250519` fails with `Vendored pin says kfxlib 20250519, but the zip contains 20260520`. 47 tier-2 tests pass.

Issue #113, opened 2026-08-22
Images preceded by a caption div are silently dropped


jloutsch:
Found by running the 77-book public-domain corpus through `convert_oeb_to_kfx`. Every image kfxgen emits is displayed and no reference dangles — but **5 of 77 books emit fewer images than their source declares**, and at least some of those are real losses, not the intentional drop from #102.

## Reproduction

`research/gutenberg-top-90/pg40739.epub` (fetch with `python research/fetch_corpus.py`) declares **8 images** in its OPF. kfxgen emits **4**.

The 8 are all unique by content hash, all referenced from spine documents. So none of the usual explanations apply — not duplicates, not unreferenced, not out-of-spine.

## The signature

| Image | Emitted | Markup |
|---|---|---|
| `fidibusz1.png` … `fidibusz4.png` | yes (4/4) | `<div class="image-center"> <img src=…>` |
| `fig1.png`, `fig2.png`, `fig3.png` | **no (0/3)** | `<div class="image-center"> <div class="caption">Fig. 1.</div> <img src=…>` |

Perfect correlation across all seven: an `<img>` that is the first child of its container survives; an `<img>` preceded by a sibling `<div class="caption">` is dropped. Nothing is logged.

The eighth image, `titelseite.jpg`, is *correctly* not emitted — it is referenced by `<link rel="icon" type="image/x-cover">`, not an `<img>`, so it is a cover declaration rather than body content.

That makes the real count 7 body images in, 4 out.

## Why nothing caught it

Two reasons, both now fixed or filed:

1. The corpus sweep called a hand-rolled pipeline ending in `generate_full_book(title, author, chapters)`, which takes no images. Every book produced zero `$164` resources regardless of source, so the sweep could not see image handling at all. Fixed on the branch that adds this corpus.
2. The golden fixtures cover the shape that works. `make_body_images` emits plain `<img>` elements; nothing in the synthetic corpus pairs a caption element with an image, so tier-3 has no way to notice.

## Scope

Across the corpus, on the shipping path:

```
77/77 books convert without error
4445 source images -> 4404 emitted -> 4578 displayed
0 orphaned resources, 0 dangling refs
72 books emit exactly the source count
 5 books emit fewer
```

The five:

| Book | Source | Emitted |
|---|---|---|
| pg1342 | 164 | 131 |
| pg37106 | 206 | 204 |
| pg1400 | 32 | 31 |
| pg40739 | 8 | 4 |
[...]

Issue #102, opened 2026-08-08
kfxgen emits four unreferenced fragments; upstream kfxlib grades it ERROR and reference files have none


jloutsch:
Found while testing something unrelated (#100). Decoding kfxgen output with the full upstream `kfxlib` from the installed KFX Input plugin:

```
ERROR: Unreferenced fragments: content_1:: $145::, img_2:: $164::,
       resource/img_2:: $417::, s0_h:: $157::
```

An Amazon-produced reference file decoded the same way produces **no such error**, so this is specific to our output rather than a quirk of the decoder or of monolithic KFX generally.

## What the four are

| Fragment | What it is |
|---|---|
| `content_1` | A `$145` content fragment — **empty**, 0 strings, 0 characters |
| `img_2` / `resource/img_2` | An image resource and its bytes, referenced by nothing |
| `s0_h` | A heading `$157` style, applied to nothing |

## Not content loss

The `$145` was the one worth checking, because an unreferenced content fragment could mean unreachable text. It does not: `content_1` holds zero strings. Verified by walking every `$259` entry's `$145` reference against the 129 content fragments in the file — 128 referenced, 1 orphaned, and the orphan is empty.

So the cost is waste and noise, not damage.

## Why it is still worth fixing

1. **Something reserves a fragment and never fills it.** An empty `content_1` alongside a populated `content_2`…`content_129` suggests the numbering or the Contents-chapter path allocates a fragment it then does not use. That is a symptom with an unexamined cause.
2. **An image is emitted that nothing displays**, along with its bytes — real weight in the container for no reading benefit.
3. **Upstream grades this ERROR, not WARN.** kfxgen's own tier-2 differential decode exists to surface exactly this kind of disagreement, and it has been reporting it all along to nobody, because that suite does not gate anything (#99).

## Reproducing

```bash
unzip -q "$HOME/Library/Preferences/calibre/plugins/KFX Input.zip" -d up
# decode with kfxlib.yj_book.YJ_Book(path).decode_book() under a logger that
# implements .warning() — kfxlib calls warning(), not warn(), and a logger
# missing it aborts the decode after two messages while looking healthy
```

That logger detail cost three false readings before the decode ran properly, so it is worth writing down for whoever picks this up.

## Scope note

The fix is probably small — find what allocates `content_1` and the orphaned image and stop emitting them when unused. The value is less in the bytes saved than in removing a standing ERROR, so that the next real one is not lost in noise.

[...]

Issue #15, opened 2026-06-28
Embed fonts in native KFX output (carry EPUB @font-face fonts through)


jloutsch:
## Summary

Add embedded-font support to the native KFX generator so fonts declared in the
source EPUB (via `@font-face` / the OEB manifest) travel inside the `.kfx` and
render on-device, instead of being dropped in favour of the installed Kindle
font.

This is tracked because lifting the "no font embedding" limitation is the
condition a Calibre maintainer set for considering kfxgen as a builtin plugin.

## Motivation

Today kfxgen deliberately does **not** embed fonts:

- `README.md` (≈L20–21, L127): "The font lives on the Kindle, not in the file —
  kfxgen does not embed font files." Framed as by-design.
- `plugin/README.md` (L38–44): "kfxgen does not embed or process fonts."

This is fine for plain prose but loses publisher typography, display faces, and
any book whose design depends on a specific embedded font (cookbooks, poetry,
technical books with custom glyphs, non-Latin scripts shipping their own font).

Important: **this is not a KFX format restriction.** KFX supports embedded fonts
(Amazon KDP output embeds them, and the kfxlib-based "KFX Output" plugin embeds
them). The limitation is specific to kfxgen's current architecture.

## Current state (why it isn't supported today)

Investigation findings, with file references:

1. **No style/CSS model.** `converter.py::extract_text_from_html` (L78–151)
   extracts chapters as `{title, text}` plain strings. It preserves inline
   `<img>` tokens but no other inline markup — no `<b>/<i>/<em>/<strong>/<span>`.
2. **Inline emphasis isn't carried yet** — separate open gap (#9). There is no
   inline "run" concept to hang per-span styling on.
3. **Styles are block-level and hardcoded.** `native_generator.py::build_fragment_157`
   (≈L922–1005) supports font-size, weight, alignment, margins, indent,
   line-height — but has **no `font-family` field**.
4. **Device font is forced.** `native_generator.py:418` hardcodes
   `override_kindle_font: False`.
5. **Fonts are skipped on extraction**; the plugin does not read Calibre's
   `embed_all_fonts` / `subset_embedded_fonts` options.
6. **Bundled `kfxlib_minimal` has no font symbols.** A grep for "font" returns
   nothing — the minimal library does not yet know font fragment/format types.

The resource-embedding mechanism itself already works: images are emitted as a
`$164` (resource metadata) + `$417` (raw binary) pair, linked by a location
name. Fonts can reuse this exact pattern.

[...]

PR #114, opened 2026-08-22
feat: make the corpus reproducible, and let the sweep see images

jloutsch:
The corpus that `BASELINE.md` measured could not be rebuilt: it recorded titles, no ids, no manifest. This makes it reproducible, and fixes the reason the sweep has never tested image handling.

## What landed

**`research/corpus_ids.json`** — 77 of the original 90 books as `{row, id, epub_images}`. The other 13 have filenames abbreviated past the point of matching a catalog entry (`69_Woordenlijst_Nederlandsche_Taal`, `71_Southern_Literature_15791895`); they need identifying by hand. No titles, not for redaction — these are public-domain Gutenberg texts — but because the id is the reproducible key and the title is not.

**`research/fetch_corpus.py`** — fetches into the already-gitignored `research/gutenberg-top-90/`, resumable, ~460 MB.

**`tests/integration/test_public_corpus.py`** — now converts through `conv.convert_oeb_to_kfx`, and checks images.

## The sweep was not testing the shipping path

`_convert` hand-rolled a reduced pipeline ending in `generate_full_book(title, author, chapters, output_path)`. That signature takes no images, so **every book produced zero `$164` resources regardless of source**. The sweep that exists to catch what unit tests miss had never seen an image, on any book, since it was written.

Measured: before, 0 of 77 books produced image resources. After, 77 of 77 do.

## Verification is the interesting part

A 200 response is not proof of a book, and three separate failures proved it:

- Mirrors answer **200 with HTML error pages** → zip validity and EPUB container are checked, not the status code.
- A 24.8 MB book arrived as a **plausible 9 MB file** when a timeout cut the transfer → Content-Length is compared against bytes written, `testzip()` runs, and downloads land via a `.part` file.
- Three ids pointed at the **wrong edition** — `pg514` returned 1 image where the baseline measured 206, because Gutenberg carries plain and illustrated editions under different ids. Caught by the expected-image-count field, resolved by downloading candidates and comparing counts: `15→2701`, `514→37106`, `79159→1260`.

That last one is why the manifest carries image counts at all: it distinguishes *a valid EPUB arrived* from *the right book arrived*.

## Also fixed: a counter that undercounts `.jfif`

pg12082 declares 436 images; 9 use `.jfif`, so an extension list without it reports 427. That is where BASELINE.md's apparent "+9 emitted" anomaly came from — it read as kfxgen inventing images. It was not: all 436 `$417` payloads carry real bytes, smallest 1,779, none empty. The generator was right and the measurement was wrong. Swept all 77 books; `.jfif` on pg12082 is the only case.

`baseline_runner.epub_image_count` shares the blind spot and is left alone — regenerating BASELINE.md is a separate change.

## New invariants

Two, matching the shape of the link checks above them:

- `$164` resources present with zero `$175` refs — the #102 defect exactly.
- `$175` refs with no `$164` behind them — the image analogue of a dangling link.

Deliberately **not** asserting source-equals-emitted: that would fail on correct code, and it would also have masked the bug below by turning it into expected noise.

Checked the guards actually fire rather than assuming: all 77 books report `image_resources > 0`, so both assertions evaluate on every book. Written against the old `_convert` they would have been vacuous.

## Results

```
77/77 convert without error
4445 source images -> 4404 emitted -> 4578 displayed
0 orphaned resources, 0 dangling refs
76/77 pass all invariants
```

The one failure is **pg64317 on the pre-existing `nav_junk` check (#60)**, not on the new ones. Its source carries a *visible* `<div class="toc"><h2>Table of Contents</h2>` in front matter — real reading content in a Gutenberg text — and `NAV_MARKERS` matches the string regardless of whether it came from a hidden landmarks nav. Left alone: narrowing a shipped-bug invariant is a decision, not a cleanup.

## Found a real bug: #113

5 of 77 books emit fewer images than their source declares, and it is not the #102 drop. In pg40739 all 8 images are unique and spine-referenced, yet 4 are emitted — an `` preceded by a sibling `<div class="caption">` is dropped, with perfect correlation across all 7 body images. Filed as #113 with the reproduction.


jloutsch:
Rebased onto `main`. This branch was accidentally cut from `feat/gate-byte-identical-goldens` rather than `main`, so it carried #111's commit and its `.github/workflows/test.yml` + `CONTRIBUTING.md` changes. Merging it would have pulled #111 in silently.

Now contains only the three corpus commits, touching three files:

```
research/corpus_ids.json                |  79 +
research/fetch_corpus.py                | 186 +
tests/integration/test_public_corpus.py |  54 +-
```

#111 and #114 are independent again and mergeable in either order. Re-verified after the rebase: 671 tests pass, and the corpus sweep is 76/77 with the one failure still being pg64317 on the pre-existing `nav_junk` check.

jloutsch:
Review addressed. The central finding was correct and important, so leading with it.

## The image assertions were satisfied by the cover alone

`native_generator.py:2169` prepends a synthetic cover chapter that emits a `$175` for the cover, and the #102 drop rule at `:2211` exempts `cover_resource_name`. Any cover-only book therefore lands on `{image_resources: 1, images_shown: 1}` regardless of body-image handling — and **46 of 77** corpus books are that shape.

Worse on illustrated books: a regression dropping every body `` ref would leave those `$164`s unreferenced, the drop rule would delete them as orphans, and the result would be cover-only counts that pass. **The check erased its own evidence.** My PR body claimed it caught "the #102 defect exactly"; it could only ever have caught a cover-side failure.

Detection has to come from the source, so the assertion now counts distinct `` targets that exist in the manifest and requires at least that many `$164` resources. Restricted to in-manifest targets so a source pointing at a file it never ships isn't blamed on the generator.

Measured before asserting: `$164 >= inline refs` holds on **72 of 77** and fires on exactly the **5** books affected by #113. No false positives — not thrown off by the cover, by images referenced twice, or by #102's legitimate drop.

**Consequence:** the sweep now reports **71/77** instead of 76/77. Every new failure is a real filed bug and the message names #113 so it doesn't read as the contributor's own change. CI is unaffected — the sweep is opt-in on `KFXGEN_CORPUS_DIR`.

## Also fixed

- **Destructive refetch** — `path.unlink()` before download only had an effect in the failure case, where it left the user with nothing. `.part` + `replace()` already handles the success case atomically.
- **`IncompleteRead` escaped the retry** — `http.client.HTTPException` isn't an `OSError`, so a truncated chunked response aborted the whole fetch with a traceback.
- **No content integrity, over a plain-http fallback mirror** — manifest now carries `sha256` per book. Verified it can fail: appending one byte to `pg84.epub` gives `sha256 167015a07a14…, manifest says 8959a5fdc05e…`, 76/77; restoring returns 77/77. This also pins editions better than image counts, which is exactly the failure three ids in this branch already hit.
- **`$175` descent** — `$181` was dead, and the old form skipped a `$175` on a parent that also had `$146` children, i.e. the nested shape `native_generator.py:1607` contemplates. Both levels counted now; corpus results unchanged, confirming that shape isn't in use yet.
- **Docstring overclaim** — it said the sweep "measures the shipping path". With `opts=None`, Stylizer construction fails so block CSS (#9) and `@font-face` (#15) degrade, and `image_optimize` is a no-op outside Calibre. Image *presence* is covered, image *processing* is not; the docstring says so now.
- **`--help`** reflowed the module docstring into one paragraph.

## Not changed

`opts=None` itself — it matches the convention in the other integration tests, and changing it is a separate decision about what the sweep should exercise.

671 tests pass.

PR #116, opened 2026-08-22
fix: images sharing a block container with a sibling were dropped (#113)

### LINKED ISSUE #113: Images preceded by a caption div are silently dropped

jloutsch:
Closes #113.

## Root cause

`extract_blocks_from_html._walk` reaches an image two ways.

A block with **no** block children is a leaf: `_walk_inline` runs over it and picks up any `` inside. A block that **does** have block children takes the container path instead, which never calls `_walk_inline` and dispatches each child to `_walk`.

Images arriving the second way were dropped. The img branch requires `not parent_is_block`, and the container passed its own blockness down, so the guard rejected them. No warning, no log line.

One-line fix: pass `parent_is_block=False` for image children. That flag only answers "has an enclosing `_walk_inline` already consumed this?" — on that path, never.

## The issue description was too narrow

Filed as "images preceded by a caption div". The caption is incidental — **any** block sibling triggers it, in either order:

| Markup | Image |
|---|---|
| `<div></div>` | kept |
| `<div><div class="caption">…</div></div>` | **dropped** |
| `<div><div class="caption">…</div></div>` | **dropped** |
| `<div><p>Cap</p></div>` | **dropped** |
| `` directly in body | kept |

## A second cause, in the harness

Two corpus books still lost their cover afterwards. `extract_cover_image` resolves `oeb_book.metadata.cover` — the manifest id from `<meta name="cover" content="…"/>` — and `EpubAsOeb` documented that attribute from the start but never populated it. Discovery fell through to the filename heuristic, so a cover not literally named `cover.*` was skipped as a cover page and then never emitted as one. `title-page.jpg` and `titelseite.jpg` both hit it.

Shim gap, not a generator defect: real Calibre supplies the attribute.

## Tests

**Eight unit cases** over the shape family — both orders, a nesting level, an exactly-once check so the fix can't double-emit what the leaf-block path already handles, and one pinning text and order around a recovered image.

**A `captioned_images` golden fixture** at tier 3. The existing `body_images` fixture wraps each `` in its own `<p>` — a leaf block, the shape that always worked — which is precisely why the golden corpus passed for months while this shipped.

Its shape assertion checks resources **and** refs. Resources alone wouldn't catch a regression: an image whose ref is lost becomes an orphan, and the #102 drop rule then deletes the resource too, so the count falls with no evidence left behind.

## The hook blocked its own documented procedure

`no-book-binaries` rejects every `*.kfx`, but `tests/fixtures/golden/expected/` **is** the tier-3 oracle and is committed on purpose — eight such files are already tracked. So the hook blocked the golden-regeneration procedure CONTRIBUTING documents. Found the hard way: the new fixture couldn't be committed.

Carve-out is exact-path, one level deep. `.epub` anywhere and `.kfx` outside that directory are still rejected — both verified.

## Verification

Regenerating goldens added `captioned_images.kfx` and **changed nothing else** — all eight existing goldens byte-identical, so the fix alters no shape that previously worked.

Mutation-tested against the layer that actually converts: reverting the fix fails the tier-3 structural diff **and** tier3_strict on the new fixture. Worth recording that the shape assertion alone does *not* catch it — it reads the committed golden, so it guards fixture rot rather than regressions. I checked that separately rather than assuming the fixture was enough.

Corpus sweep **71/77 → 74/77**. Recovered: pg1342 (33 images), pg1400, pg40739, pg42486.

## Still failing, deliberately not fixed here

- **pg1400, pg37106** — 1 and 2 images short from a **third** cause, with no dangling refs and every emitted resource referenced. Different root cause, filing separately rather than bundling.
- **pg64317** — the pre-existing `nav_junk` false positive on a visible front-matter TOC heading.

687 tests pass.


jloutsch:
All five findings addressed. The medium one was correct and would have let the regression back in.

## The shim's cover parsing had no test at all

Confirmed by the reviewer's method: deleting the entire `<meta name="cover">` scan leaves all 687 tests and every golden passing. The reason is that `EpubBuilder.set_cover` hardcodes `id="cover-image"`, which the filename heuristic matches regardless of href — so no fixture built with it can distinguish the metadata path from the fallback.

Three tests now cover it, on a hand-rolled EPUB whose cover is `plate` / `title-page.jpg`. Neither the id nor the href contains "cover", so the heuristic provably cannot find it. That is the exact corpus shape (`title-page.jpg`, `titelseite.jpg`) this PR set out to fix.

## EPUB 3 covers were still missed

Only the legacy EPUB 2 form was read. EPUB 3 declares the cover as `properties="cover-image"` on the manifest item and often omits the legacy meta entirely — identical failure mode for any such producer. The Gutenberg corpus emits the legacy form, which is why the sweep passed without it. Both are read now, with a test.

## The dead guard is gone, not documented

`parent_is_block` could never be true after the fix, and if a future call site ever passed `True` the image would fall through, produce no block, and log nothing — reintroducing #113 invisibly. Removed the guard and the parameter rather than leaving the trap in place.

The comment justifying it was also factually wrong, as noted: it claimed the leaf-block branch is the only caller of `_walk_inline`, when the container branch calls it ten lines below. The conclusion held; the stated reason did not. Since that comment *was* the justification, it's now gone along with the guard.

## The `== 2` was measuring identical bytes

Both fixture images used `_MINIMAL_JPEG`, so a future content-dedupe would drop the count to 1 and the test would confidently report a #113 regression that had not happened. The second image is now a PNG, so the fixture also covers a mixed-format container.

Two attempts worth recording. A 1×1 PNG is silently discarded — `extract_images_from_oeb` skips any image of 100 bytes or fewer (`converter.py:1171`) — and a solid-colour PNG compresses under that floor at any dimension. The committed payload is a 16×16 incompressible gradient, 819 bytes, all chunk CRCs verified.

## Verification

Regenerating goldens changed **only** `captioned_images`; the other eight stayed byte-identical, which is what confirms removing the guard and parameter alters no behaviour.

Re-ran the mutation test after the refactor, since the earlier one tested code that no longer exists: disabling the img branch fails 10 tests, including the tier-3 structural and byte-identical diffs.

690 tests pass. Corpus 74/77 — the three remaining are #117 (illustrations inside a discarded Contents section, filed separately with the trace) and the pre-existing `nav_junk` false positive on pg64317.

Issue #120, opened 2026-08-22
kfxgen never emits $601 render mode; Amazon marks decorative images inline


jloutsch:
Found while disproving #118, by comparing kfxgen's output against Kindle Previewer 3.106 on the same book (`research/gutenberg-top-90/pg40739.epub`). This is the one genuine conformance gap that comparison turned up.

## The difference

Both pipelines emit 8 `$164` image resources and 8 `$271` image entries. Amazon sets `$601` on four of them; kfxgen sets it on none.

```
kfxgen entry shapes:   x7  [$155, $157, $159, $175, $584]
                       x1  [$155, $157, $159, $175, $584, $790]

Amazon entry shapes:   x4  [$155, $157, $159, $175, $584, $601]   ($601 = $283)
                       x3  [$155, $157, $159, $175, $584]
                       x1  [$155, $157, $159, $175]
```

`$601` is the render property. `$601 == "$283"` maps to `-kfx-render: inline` in upstream's decoder (`yj_to_epub_content.py:1312`), selecting inline flow rather than block for the image.

## Which images, and why it is not arbitrary

Mapping Amazon's resources back by dimension:

| `$601` | Images | Source markup |
|---|---|---|
| `$283` (inline) | 512×206, 512×204, 512×202, 512×197 — the four decorative *Fidibus* images | `<div class="image-center"><img></div>` — image alone in its container |
| absent (block) | 551×197, 544×188, 553×191 — the three captioned figures | `<div class="image-center"><div class="caption">Fig. N.</div><img></div>` |
| absent (block) | 462×700 — the title page | cover wrapper |

Amazon marks an image inline when it stands alone in its container, and leaves it block when it sits with a caption. That is exactly the distinction `extract_blocks_from_html._walk` already makes internally — leaf block versus container with block children, the same split #113 turned on — but kfxgen discards it after extraction instead of recording it in the output.

## Why this might matter

Untested, and worth saying so plainly: no device check has been done, and the practical effect of `-kfx-render: inline` on a Kindle is unknown to me. It plausibly changes spacing around decorative images — whether they sit in the text flow or break it — which on a page of body text is visible.

It is also the kind of difference that is invisible to every check kfxgen currently has. Tier-2 decodes without raising, tier-3 compares against our own goldens, and the corpus sweep counts images. None of them compare against Amazon.

## Suggested approach

[...]

Issue #118, opened 2026-08-22
Images are not tappable: kfxgen never emits the zoomable plugin resource


jloutsch:
Found on device while verifying #116. Images render, but tapping one does nothing — no enlarge, no full-screen view. Observed on both test books.

Not a regression from #116. kfxgen has never emitted the structure this needs.

## What kfxgen emits

`native_generator.py:460` picks the resource format from the image bytes:

```python
format_sym = "$285" if image_format == "jpeg" else "$284"
```

So every image becomes a plain `$164` resource with `$161` set to `$284` (PNG) or `$285` (JPEG), referenced from a `$259` entry via `$175`. That is a complete, valid image — and a static one.

## What tap-to-zoom actually requires

Upstream `kfxlib` shows the two paths side by side when converting KFX back to EPUB (`yj_to_epub_misc.py:517`):

```python
elif res.format == "$284":
    content_elem.tag = "img"          # plain image, nothing tappable
    ...
else:
    manifest_ = IonText(...).deserialize_annotated_value(res.raw_media.decode("utf-8"), ...)
    plugin_type = manifest_.get_annotation()
    ...
    elif plugin_type == "zoomable":
        content_elem.tag = "img"
        content_elem.set("src", ... manifest["facets"]["media"]["uri"] ...)
```

A zoomable image is a **plugin resource**: its payload is not image bytes at all but an Ion *text* manifest, annotated with the plugin type, whose `facets.media.uri` points at the real image resource. The same mechanism carries `audio`, `button`, `hyperlink` and `webview`.

So the shape is roughly:

- the image resource itself, as today
- a second resource whose raw media is an Ion-text manifest annotated `zoomable`, referencing the first
- the `$259` entry pointing at the plugin resource rather than the image

Exact field names beyond `facets.media.uri` still need reading out of a real Amazon-produced file; the decoder above only shows what it consumes.

## Why it matters

Illustrated non-fiction is where this bites. #116 just recovered 33 plates in one corpus book and three schematic figures in another — diagrams that are legible only when enlarged on a 6-inch screen. Getting the image into the file and leaving it untappable solves half the problem.

It also affects every image kfxgen has ever emitted, not just the recovered ones.

## Verification path

`research/kfx-format-baseline/` already decodes Amazon-engine output through full `kfxlib`. A Previewer conversion of a book with a zoomable figure would show the real manifest, which is the cheapest way to get the field names right rather than guessing them.

[...]

PR #121, opened 2026-08-22
test: pin the image-entry contract verified against Amazon's output (#118)

jloutsch:
#118 asked for a zoomable plugin resource so images would respond to tap. **I did not implement it — the premise was wrong**, and this PR records why plus a test for the shape that is actually correct.

## What the measurement showed

Converted the same book with Kindle Previewer 3.106 and decoded the KPF with full upstream `kfxlib`:

| `pg40739.epub` | kfxgen | Previewer 3.106 |
|---|---|---|
| `$164` image resources | 8 | 8 |
| Plugin resources | 0 | **0** |
| `$274` plugin content | none | **none** |
| Image content type | `$271` | `$271` |

Amazon's content types in that file: `$269`, `$270`, `$279`, `$596`, `$271`, `$278`, `$454`. No `$274`.

The zoomable path in `yj_to_epub_misc.py` is real, but it is reached only from `content_type == "$274"`, which EPUB conversion never produces — it is presumably for comics with region magnification or books authored in Kindle Create. Emitting it here would mean shipping structure the reference implementation never emits, unverifiable except by another device round-trip. That is divergence, not conformance.

#118 is closed as won't-do with the measurements recorded on it.

## What this PR adds

A tier-3 assertion pinning the verified shape: `$271` content type, every referenced resource present, every `$164` backed by a `$417` payload.

The docstring carries the comparison, because the failure mode here is not a bug — it is someone reading the tap-to-zoom path in a year, assuming it applies to ordinary images, and implementing it. The test makes that a deliberate decision rather than a drift.

Verified the assertion discriminates rather than assuming: flipping a golden's image entries to `$274` is detected.

## Also found, filed separately as #120

The comparison surfaced one genuine conformance gap. Amazon sets `$601` (`-kfx-render: inline`) on 4 of its 8 image entries; kfxgen sets it on none. Mapping by dimension shows the rule:

| `$601` | Images | Source markup |
|---|---|---|
| `$283` inline | four decorative images, 512×~200 | `<div></div>` — image alone |
| absent, block | three captioned figures + title page | `<div><div class="caption">…</div></div>` |

Amazon marks an image inline when it stands alone in its container. That is the same distinction `_walk` already makes internally — the one #113 turned on — but kfxgen discards it after extraction instead of recording it. Filed rather than fixed here: one book is one data point, and it should be confirmed on two or three more before implementing.

691 tests pass.

### DIFF
--- a/tests/integration/test_golden_corpus.py
+++ b/tests/integration/test_golden_corpus.py
@@ -493,3 +493,59 @@ def test_fixture_long_chapter_shape(tmp_path):
             )
             refs += 1
     assert refs > 0, "no text entries carried a $145 reference"
+
+
+@pytest.mark.tier3
+@pytest.mark.integration
+def test_image_entries_match_amazons_shape():
+    """Images use the plain `$271` content type, not the `$274` plugin type.
+
+    Recorded from a direct comparison rather than inference. Kindle Previewer
+    3.106 was run over `pg40739.epub` and the resulting KPF decoded with full
+    upstream `kfxlib`; both pipelines produced 8 `$164` resources, 8 `$271`
+    image entries, and **zero** plugin resources. Amazon's content types in
+    that file were `$269`, `$270`, `$279`, `$596`, `$271`, `$278`, `$454` —
+    no `$274` anywhere.
+
+    That matters because `$274` routes to `process_plugin` in upstream's
+    decoder (`yj_to_epub_misc.py:534`), which is where the "zoomable" plugin
+    lives. #118 proposed emitting it so images would respond to tap; the
+    comparison showed Amazon never emits it for EPUB conversion either, so
+    doing that would be divergence from the reference rather than conformance.
+    This test pins the shape that was verified, so a future change toward `$274`
+    has to be a deliberate decision rather than a drift.
+
+    Every referenced resource must also have a payload: a `$164` with no `$417`
+    is a declared image with no bytes behind it.
+    """
+    frags = load_fragments(EXPECTED_DIR / "captioned_images.kfx")
+    resources = {str(f.fid) for f in by_type(frags, "$164")}
+    payloads = {str(f.fid).split("/")[-1] for f in by_type(frags, "$417")}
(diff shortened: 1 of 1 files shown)


jloutsch:
All four addressed. The second finding was the important one and it is the third time this exact trap has caught me today.

## It read the committed golden

So a generator regression would leave the golden untouched and the test would pass while the product was broken — contradicting its own docstring and the policy this file states three times.

The reviewer's scenario is worse than I would have guessed: the whole **default suite** stays green. `test_golden_structural_diff` compares only top-level key names and `$159` is nested, so it sees nothing. Only `tier3_strict` catches it, and its remedy line says "regenerate goldens" — after which the evidence is gone.

Now built with `_build_fresh`, matching the neighbouring shape tests.

**Verified by mutating the generator, not the data.** Flipping `native_generator.py:1524` from `$271` to `$274`:

```
before this change:  691 passed        <- the regression sailed through
after:               1 failed, 690 passed
```

## The payload check matched basenames, not the link

`$164` and `$417` are paired through `$165`, and the location name is documented as free-form (`"cover_img_raw"`). Comparing `$164` fids against `$417` basenames only worked because they happen to share a suffix today.

Demonstrated the failure mode concretely — a *valid* file using the documented naming:

```
$164 fid = cover_img, $165 = cover_img_raw, $417 fid = cover_img_raw

new ($165 link)      -> passes: True
old (basename match) -> passes: False    <- rejects a correct file
```

Now resolves `$165` against `$417` fids.

## The walk was one level deep and `$146`-only

An image entry nested deeper — the container case #113 was about — was never content-type checked, and a storyline using `$181` would read as "no entries" and trip the fixture-rot assert as a false failure. Replaced with a recursive struct walk.

## The docstring cited an upstream line number

In a file that is not vendored here and not pinned to a version — the part most likely to be wrong in a year, in a docstring whose whole purpose is to be read then. Now names the behaviour and pins the kfxlib version.

691 tests pass.

Issue #67, opened 2026-08-01
Subscript baseline shift is an unverified guess with no reference or device test


jloutsch:
## What shipped

`SUBSCRIPT_SHIFT = ("-20", "$314")` in `plugin/kfxgen/native_generator.py` — a
negative `$31` baseline shift applied to `<sub>` and to CSS
`vertical-align: sub`/`text-bottom`.

## Why it is debt

The superscript value beside it is grounded: `$31: 35%` with `$16: 0.75rem` is
copied from Amazon-produced noteref styles and is device-verified.

The subscript value is not. **No reference KFX in the corpus contains a negative
`$31`**, no test book exercises a subscript, and it has never been rendered on
hardware. −20% was chosen because CSS `sub` sits roughly a fifth of the font
size below the baseline — a reasoned guess, not an observation.

Unknowns:

- whether a negative `$31` is even valid in KFX (no reference uses one)
- whether the reader clamps, ignores, or mis-renders it
- whether the magnitude is right

## Options

1. Verify on device with a book containing chemical formulae or similar, and
   record the result.
2. Find a reference KFX that uses subscript and copy its encoding.
3. Drop `<sub>`/`vertical-align: sub` handling until one of the above, so a
   guessed value is not shipping.

Option 3 is the honest default if this stays unverified — rendering a subscript
wrongly is worse than leaving it at the baseline, which is what happened before
and which nobody reported as a bug.

## Related

Superscript support and the `$31` mechanism landed in #52.

jloutsch:
Device-verified ✅

Sideloaded `kfxgen-issue67-subscript-test.kfx` (title *ZZ-#67 SUB+SUPER TEST*). Both render correctly on a physical Kindle:

| | value | route covered |
|---|---|---|
| superscript | `$31 = 35%`, `$16 = 0.75` | `<sup>` tag and CSS `vertical-align: super` |
| subscript | `$31 = -20%`, `$16 = 0.75` | `<sub>` tag and CSS `vertical-align: sub` |

The test book pairs each form with an unmarked baseline line, so the comparison was direct rather than an impression.

This resolves the open question in this issue: a **negative `$31` is valid** and renders as expected, despite no reference KFX in the corpus using one. The −20% value was reasoned from CSS convention rather than copied, and it holds up. Both values are now device-verified, and #68 made them env-tunable (`KFXGEN_SUBSCRIPT_SHIFT_PCT`) should a different device disagree.

No unverified rendering constants remain in the codebase.

PR #122, opened 2026-08-22
fix: give subscripts their own font-size variable (#115)

### LINKED ISSUE #115: Subscript font size can only be tuned through a variable named for superscript
Noticed while reviewing #111. Not a rendering defect — both values are device-verified — but the override mechanism #68 built is asymmetric in a way its own naming hides, and one code comment now contradicts the device record.

## The coupling

`plugin/kfxgen/native_generator.py:71`, inside `subscript_metrics()`:

```python
size = _env_number("KFXGEN_SUPERSCRIPT_FONT_SIZE", SUPERSCRIPT_FONT_SIZE)
pct  = _env_number("KFXGEN_SUBSCRIPT_SHIFT_PCT", SUBSCRIPT_SHIFT_PCT)
```

The *shift* is per-type; the *size* is shared, under the superscript name. Setting `KFXGEN_SUPERSCRIPT_FONT_SIZE` to fix a device that renders superscripts too small also resizes every subscript, and nothing in the variable name says so. There is no `SUBSCRIPT_FONT_SIZE` constant and no `KFXGEN_SUBSCRIPT_FONT_SIZE`.

## Why this is minor, stated plainly

The shared default is right. #67 closed device-verified — the sideload of *ZZ-#67 SUB+SUPER TEST* confirmed superscript at `$31 = 35%, $16 = 0.75` and subscript at `$31 = -20%, $16 = 0.75` on a physical Kindle. Both were checked at the same font size, so sharing the constant reflects what the hardware showed.

What is missing is the escape hatch. `superscript_metrics()`'s docstring gives the rationale for these overrides existing at all:

> Exposing them means a Kindle that renders superscripts too high or too small can be corrected without a rebuild. (#68)

A Kindle that renders *subscripts* too small has no equivalent fix. Correcting it drags superscripts along.

## Two smaller things in the same place

**A stale comment contradicting the device record.** `native_generator.py:40`:

```python
SUBSCRIPT_SHIFT_PCT = -20  # -20% (see #67 — not device-verified)
```

#67 is closed as completed with a device-verification comment, and `docs/kfx-generation-explained.md:273` correctly says the defaults are "all device-confirmed". The code comment is the only place still saying otherwise. It is the mirror image of the usual problem here: a record claiming *less* confidence than the evidence supports, which invites someone to re-litigate a value that was actually checked on hardware.

**The coupling is untested.** `tests/unit/test_native_generator.py:2075` exercises `KFXGEN_SUBSCRIPT_SHIFT_PCT` and asserts the default size of `0.75`, but never sets a font-size variable, so nothing pins whether subscript size follows the superscript variable, a subscript variable, or neither. Whichever way this is resolved, that behaviour should be asserted.

## Options

1. **Document the coupling and leave the code alone.** The docs table at `docs/kfx-generation-explained.md:277` already labels the row generically as "Font size"; adding "(applies to subscripts too)" makes the shared scope explicit at essentially zero risk. Cheapest, and does not touch device-verified output.
2. **Add `KFXGEN_SUBSCRIPT_FONT_SIZE`, defaulting to the superscript value.** Preserves current behaviour and output bytes for anyone who sets nothing, and gives subscripts their own dial. Slightly more surface.
3. Rename the shared variable — rejected: it is released (5.x, see CHANGELOG:183) and breaking it buys nothing.

Either way, fix the stale `#67` comment and add the missing test.

## Priority

Low. No output changes today, no rendering regression, and both values are hardware-confirmed. Filed because the naming actively misleads: someone tuning superscripts on a device will move subscripts without noticing, and the one comment that would warn them says the opposite of what the device test found.

jloutsch:
Closes #115.

## The gap

`subscript_metrics()` read `KFXGEN_SUPERSCRIPT_FONT_SIZE`. The *shift* was per-type but the *size* was shared, under a name that said superscript — so a device rendering subscripts too small could only be corrected by resizing every superscript too, with nothing in the variable name to warn you.

That undercuts the stated reason these overrides exist. `superscript_metrics` promises "a Kindle that renders superscripts too high or too small can be corrected without a rebuild"; subscripts had no equivalent.

## The fix

`KFXGEN_SUBSCRIPT_FONT_SIZE`, falling back to `KFXGEN_SUPERSCRIPT_FONT_SIZE` before the built-in default.

That ordering is the whole design. Setting only the superscript variable still moves both — the behaviour that shipped in 5.x — so no existing setup changes under anyone. Narrowing it to superscripts only would have silently altered output for current users.

**No output change.** Both defaults are 0.75, and `tier3_strict` passes with no golden moved.

## A stale comment pointing the wrong way

The constant block claimed the subscript shift is "NOT device-verified". #67 was closed by a physical sideload of *ZZ-#67 SUB+SUPER TEST*, which rendered both forms correctly over both the tag route (`<sup>`/`<sub>`) and the CSS route (`vertical-align`).

A record claiming *less* confidence than the evidence supports is its own hazard — it invites someone to re-litigate a value the hardware already settled. Corrected, with the device evidence cited inline.

## The tests that were missing

The existing subscript test set only the shift variable, so **nothing pinned which variable subscript size followed** — the coupling was never asserted in either direction. Four tests now cover:

- the new dial works, and does not drag superscripts along
- the superscript variable still moves both when no subscript variable is set
- the specific variable wins when both are set
- both defaults equal the device-verified values

Verified non-vacuous: reverting `subscript_metrics` to read only the superscript variable fails 2 of them.

695 tests pass.

## Found while investigating, filed separately

Comparing this book against Kindle Previewer 3.106 output shows a difference in how the two express raised text. Amazon uses `$44` baseline-style — the semantic enum, `$370` super / `$371` sub — combined with the source's own font size:

```
x2  $44=$370  $16=0.8
x1  $44=$370  $16=0.936
x1  $44=$370  $16=0.833333
x1  $44=$371  $16=0.694444
```

kfxgen uses `$31`, a numeric percentage shift, with a fixed 0.75. Amazon emitted **zero** `$31` in that book. Both render — kfxgen's approach is device-verified — but the idioms differ, and Amazon's lets the device decide the offset while preserving publisher font sizes. Filed rather than bundled here.


jloutsch:
All three applicable findings addressed, plus a branching error the review surfaced indirectly.

## The branch was cut from the wrong place

The review cited `test_golden_corpus.py:498`, which belongs to #121 — that file should not have been in this PR at all. `fix/115-subscript-font-size` was branched off `test/pin-image-entry-contract` instead of `main`, so it carried #121's two commits.

Rebased onto `main`. Now one commit, three files:

```
docs/kfx-generation-explained.md    | 10 +
plugin/kfxgen/native_generator.py   | 40 +-
tests/unit/test_native_generator.py | 63 +
```

Finding 4 was about `_iter_structs` / `_count_image_resource_entries`, which live in #121 — no longer in scope here, and worth carrying over there.

Second time today I have made this mistake (#114 was cut from #111 the same way).

## `SUBSCRIPT_FONT_SIZE` was an alias

`= SUPERSCRIPT_FONT_SIZE` contradicted the comment directly above it. Editing the superscript constant on a device report would have silently moved the subscript default too — the exact coupling this PR breaks, reintroduced one level down. Now a literal `0.75`, with the reason stated.

## The fallback was evaluated eagerly

Passing the superscript lookup as `_env_number`'s default argument evaluates it even when the subscript variable is set and wins. A malformed superscript value would log "ignoring invalid ..." on every call, and `subscript_metrics()` runs once per emphasis run — so a book with many subscripts buries the log in warnings about a value that changed nothing.

Resolves lazily now. Verified: malformed superscript variable plus a valid subscript variable produces **0 warnings over 5 calls**, and the subscript variable still wins.

## Tests read the real environment

Reproduced exactly as reported — and it is worse than the new tests, because two of the three failures are pre-existing:

```
KFXGEN_SUPERSCRIPT_FONT_SIZE=0.9 pytest -k "subscript or superscript"
→ 3 failed
```

These variables are *meant* to be exported for a conversion; that is the documented way to correct a device. So a maintainer following the documented workflow saw failures unrelated to their change.

Fixed with an autouse fixture in `tests/conftest.py` clearing `KFXGEN_*`, repo-wide rather than local: the leak was never specific to these tests — any test asserting a default was exposed to whatever the shell carried. The same command now passes, and the suite is unaffected on a clean environment.

Noted in the fixture docstring that `KFXGEN_MAX_DECODE_SIZE` is read at import time so clearing it there has no effect; it stays in the sweep so the rule remains "no `KFXGEN_*` reaches a test".

694 tests pass.

jloutsch:
Added a subscript golden fixture. The gap was real, and larger than "missing coverage" — the corpus was completely blind to the code path this PR changes.

## Measured, not assumed

On `main`, changing `SUBSCRIPT_SHIFT_PCT` from -20 to -25 and running the entire tier-3 corpus:

```
36 passed, 0 failures
```

Nothing noticed. With `sub_super_marks` in place, the same mutation fails 2 tests, and mutating `SUBSCRIPT_FONT_SIZE` from 0.75 to 0.7 also fails 2.

## Why it was blind

`publisher_structure` exercises `<sup>`, so superscript changes moved bytes and were caught. Nothing reached `subscript_metrics` — no fixture emitted a `<sub>`, and no golden contained a `-20` shift. So the byte-identity claim for this PR rested on a unit test, not a golden diff.

That is the same shape as the gap that let #113 sit for months: the corpus covered the variant that worked and not the one that did not.

## The fixture

`sub_super_marks` emits both forms and locks both: `$31` +35% / -20% with `$16` 0.75 on each. Both are device-verified (#67), so it pins evidence rather than a preference.

Tag route only, deliberately — the CSS route resolves through Calibre's Stylizer, which golden generation runs without, so it would silently emit no span. That is the trap `make_publisher_structure` already documents; the CSS route stays unit-tested with an injected stylizer.

The shape assertion builds fresh rather than reading the committed golden, so a metrics change cannot pass by leaving the golden untouched.

All eight existing goldens unchanged — only the new file was created. 702 tests pass.

## Merge readiness

- **#121** — test-only, one file, no production code. Nothing to test.
- **#122** — no device test warranted. Defaults are unchanged, rendering only differs if someone sets the new variable, and the values it defaults to are the ones #67 verified on hardware. The subscript path now has real tier-3 coverage rather than inferred coverage.

Issue #109, opened 2026-08-21
The device tier gates releases but leaves no per-change record


jloutsch:
Noticed while closing #91 and reviewing what pre-merge review can reasonably exist on this repo. Tier 4 is the only oracle that proves a KFX renders, and it is the one tier with no structured record.

## What the repo says tier 4 is

- `pytest.ini:39` declares the marker: `device: requires physical Kindle device (manual, gates release tags only)`.
- `pytest.ini:17` excludes it by default; `:33` says release tags require manual `device` sign-off.
- `CONTRIBUTING.md:97` defines tier 4 as manual verification on a physical Kindle (Paperwhite/Oasis/Voyage).
- `CONTRIBUTING.md:110` — release notes "must reference which devices were tested".
- `CONTRIBUTING.md:135` shows `@pytest.mark.device` as the example to copy.
- `CONTRIBUTING.md:258` — for changes touching `$164`/`$417`, "real-device validation (tier 4) is **mandatory before merging**".

## What is actually there

**Zero tests carry the marker.** `grep -rn "mark\.device" tests/` returns nothing. The marker is declared, excluded by default, documented as the example, and used by nothing — so `pytest -m device` is a no-op that exits successfully. A gate that passes because it is empty looks identical to one that passes because it ran.

**The record exists, but only as prose in CHANGELOG.** This part is better than it first appears and worth stating fairly — there are 9 device references, some of them good:

- A device-behavior table for the thumbnail work (Paperwhite vs Voyage, with the firmware limitation called out).
- "**Device-verified.** Sideloading a controlled pair (identical build and book, differing only by `KFXGEN_IMAGE_MAX_BYTES`, with distinct titles and ASINs…)" — the methodology is recorded, not just the outcome.
- "confirmed by sideload, the only render test available for raw `.kfx`".

So the practice is real and sometimes carefully documented. The gaps are narrower than "nothing is recorded":

1. **Version granularity, not change granularity.** The CHANGELOG entry covers a release. `CONTRIBUTING.md:258` makes device validation mandatory *before merging* a specific class of change, and nothing at merge time records whether it happened for that change. Asking "was this commit device-checked?" is not answerable.
2. **Device model named inconsistently.** Sometimes "a physical Kindle" (unnamed), sometimes a full model table. `CONTRIBUTING.md:110` asks for the former to be the latter.
[...]

PR #125, opened 2026-08-22
feat: opt-in $44 baseline-style encoding for the #123 device comparison

### LINKED ISSUE #123: Raised text uses $31 numeric shift; Amazon uses $44 baseline-style

jloutsch:
Builds on #124 (which pins the current choice). This makes the comparison #123 needs actually runnable.

## Why an opt-in rather than a switch

#123 established that Amazon encodes raised text with `$44` baseline-style and never with `$31`, measured over four books. kfxgen emits `$31`, which #67 verified on hardware.

What is unknown is whether `$44` renders **as well or better**. Amazon preferring it is evidence about Amazon's toolchain, not about whether a Kindle draws kfxgen's output correctly — and the device already said it does. Switching blind trades a verified encoding for an unverified one on a conformance argument.

So: build both from one source, read them side by side, then decide.

## The switch

`KFXGEN_RAISED_TEXT_ENCODING=baseline-style` emits `$44` (`$370` super, `$371` sub). Anything else keeps `$31`. Default unchanged, no golden moved, `tier3_strict` clean.

Direction comes from the sign of the shift that would otherwise have been emitted, so both source routes (`<sup>`/`<sub>` tags and CSS `vertical-align`) and both font-size overrides keep working with no second code path.

**Fails closed.** `1`, `true`, `yes`, `shift` all keep the default — only the explicit `baseline-style` / `baseline_style` / `44` switches it. A typo or a generic truthy value must not silently select the unverified encoding.

## This is meant to be deleted

An experiment with an exit condition, marked as such at the constant: once the sideload says which encoding kfxgen should emit, the switch and its branch go. It should not linger as a permanent second path — that is exactly the "experimental code paths left in production" shape.

## Tests

Both paths, 13 cases: default emits `$31` and no `$44`; the flag emits `$44` with correct direction and never both; five unrecognised values pinned to the default.

## Comparison artifacts

Built on the maintainer's Desktop from one source in both encodings:

```
ZZ-123-supsub-A-shift-current.kfx        (current, $31 ×2)
ZZ-123-supsub-B-baselinestyle-123.kfx    (proposed, $44 = $370/$371)
```

Purpose-made rather than a corpus book. pg25851's 18 raised runs are all asterisks — hard to judge a few percent of baseline shift by. This one has `E = mc²`, `H₂O`, `CO₂`, ordinals (1ˢᵗ, 2ⁿᵈ), indices, and a mixed `x₁² + x₂² = r²`, plus a "normal text → H₂O → normal" line as a baseline reference.

Verified the two differ only in encoding: A has 2 `$31` styles and no `$44`; B has `['$370', '$371']` and no `$31`.

Sideloading both and comparing is what closes #123. Worth doing on the Paperwhite (5.19.2) as well as the Oasis, since that also exercises current firmware — see #109.

703 existing tests plus 13 new pass.

### DIFF
--- a/plugin/kfxgen/native_generator.py
+++ b/plugin/kfxgen/native_generator.py
@@ -55,6 +55,31 @@ SUBSCRIPT_FONT_SIZE = 0.75
 SUBSCRIPT_SHIFT_PCT = -20  # -20% (device-verified, #67)
 
 
+#: Opt-in encoding for raised text, for the device comparison #123 needs.
+#:
+#: EXPERIMENTAL AND TEMPORARY. It exists so the two encodings can be built from
+#: the same source and read side by side on hardware; once that comparison
+#: settles which one kfxgen should emit, this switch and the branch it feeds
+#: should be deleted rather than left as a permanent second code path.
+RAISED_TEXT_ENCODING_ENV = "KFXGEN_RAISED_TEXT_ENCODING"
+
+
+def _use_baseline_style():
+    """True when raised text should use `$44` baseline-style instead of `$31`.
+
+    Amazon emits `$44` (`$370` super, `$371` sub) and never `$31` — measured
+    over four books via Kindle Previewer 3.106 (#123). kfxgen emits `$31`, and
+    that form is device-verified (#67), so the default does not move. What is
+    unknown is whether `$44` renders *as well or better*, and nothing but a
+    sideload answers it.
+    """
+    return os.environ.get(RAISED_TEXT_ENCODING_ENV, "").strip().lower() in {
+        "baseline-style",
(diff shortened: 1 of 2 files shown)


jloutsch:
Superseded by #126. The opt-in flag existed to make the device comparison possible; the comparison is done and `$44` is adopted, so the flag is removed rather than inverted — carrying it would be the experimental-path-in-production shape its own comment warned against. The comparison artifacts it produced did their job. Closing unmerged.

PR #126, opened 2026-08-22
feat: adopt $44 baseline-style for raised text (#123)

### LINKED ISSUE #123: Raised text uses $31 numeric shift; Amazon uses $44 baseline-style

jloutsch:
Closes #123. Supersedes #124 and #125 — see below.

## The device answered it

The comparison pair from #125, built from one source and read on a **Paperwhite running 5.19.2**: both encodings displayed superscript and subscript as expected. `$44` is no longer the unverified option, which removes the only argument for keeping `$31`.

Worth noting that this is **current firmware**. Every prior device result in this repo came from an Oasis on 5.18.2, which is terminal — 5.19.2 is the firmware with known sideload rendering changes and had never been tested here (#109).

## Adopted on the delegation argument, not conformance

A percentage is a guess at the right offset. The enum states the intent and lets the reader compute the offset from its own font metrics, which is more likely to hold on hardware and firmware nobody here can test.

That Amazon emits `$44` exclusively — four books, `$31` zero times — is evidence it reached the same conclusion. It isn't the reason by itself; conformance with a reference is a proxy for correctness, not correctness.

## The shift percentages are removed, not left inert

Under `$44` the device chooses the offset, so `SUPERSCRIPT_SHIFT_PCT`, `SUBSCRIPT_SHIFT_PCT` and their `KFXGEN_*_SHIFT_PCT` overrides have nothing to act on.

Keeping them would leave **documented knobs that silently do nothing** — the exact failure mode this codebase keeps finding elsewhere (#99, #109, and the pre-push gate in #107). Setting either now logs a warning naming the replacement, with a test asserting it appears. Being ignored quietly was the one outcome not on the table.

Font sizes are untouched and stay overridable, including #115's fallback.

## Golden churn, per CONTRIBUTING

**Two of eleven** goldens changed, and only the two containing raised text:

| Golden | Change |
|---|---|
| `sub_super_marks` | `$157` loses `$31`, gains `$44` = `$370` / `$371` |
| `publisher_structure` | same, `$370` only |

In both, `$31` (a struct of `$307` magnitude + `$306` unit) is replaced by `$44` (a bare symbol). `$16` is unchanged at 0.75.

The other **nine goldens are byte-identical**, which is the evidence that nothing outside raised text moved.

## Supersedes the two open PRs

- **#124** pinned `$31` as the choice. That decision is now reversed, so its assertion is inverted here instead — the golden shape test asserts `$44` present and `$31` absent.
- **#125** added the opt-in flag to make this comparison possible. The comparison is done, so the flag is gone rather than inverted. Carrying it would be the experimental-path-in-production shape its own comment warned about.

Both can be closed unmerged.

## Verification

705 tests pass, `tier3_strict` clean. Tests updated to the new contract rather than deleted — the builder emits `$44` and not `$31`, super yields `$370` and sub `$371`, composition-with-italic still holds, and the retired variables warn.

A fresh default build now emits exactly what file B contained, so what ships is what was read on the device.


jloutsch:
## Verified on both devices — the review's remaining concern is closed

| Device | Firmware | Status | `$44` result |
|---|---|---|---|
| Paperwhite | 5.19.2 | current | superscript and subscript display correctly |
| Oasis, 10th gen | 5.18.2 | terminal | superscript and subscript display correctly |

The review's fifth finding was the one that mattered for merging: `$44` had been confirmed only on 5.19.2, while the encoding it replaces was verified on 5.18.2, and `$31` is removed with no fallback. If an older reader ignored the enum, superscripts would degrade to small-but-unraised text and nothing in the suite could detect it — goldens and tier-2 decode pass either way.

That is now tested rather than argued. The Oasis matters specifically because it is on terminal firmware: it stands for every device that will never update, so a regression there would have been permanent for that population.

The Oasis file also carried a descender/ascender case (`pg<sub>j</sub>qy`, `Ah<sup>l</sup>Rk` with a line directly beneath), since delegating the offset to the device makes clipping or collision with adjacent lines the plausible failure mode — not something `H₂O` on its own would show.

## Evidence summary for this change

- **Correctness of the encoding**: `$44` = `-kfx-baseline-style`, `$370` super / `$371` sub, confirmed against upstream kfxlib's property table.
- **Conformance**: Amazon emits `$44` exclusively across four books via Kindle Previewer 3.106; `$31` appears zero times.
- **Rendering**: correct on two devices spanning two firmwares, one of them current.
- **Blast radius**: 2 of 11 goldens changed, both containing raised text; the other 9 byte-identical.
- **Retired config**: warns once per book, including books with no raised text.

707 tests pass, `tier3_strict` clean, all four checks green.

Ready to merge as far as I can tell.

jloutsch:
Device record completed — the Paperwhite is an **11th generation** (2021).

Full verification for this change:

| Device | Firmware | Status | `$44` |
|---|---|---|---|
| Oasis, 10th generation (2019) | 5.18.2 | terminal | superscript and subscript display correctly |
| Paperwhite, 11th generation (2021) | 5.19.2 | current | superscript and subscript display correctly |

Two hardware generations, two firmware lines, one of them current and one that will never update again. Recorded on #109 so the next person can tell what was actually tested rather than reading "device-verified" and guessing.

Issue #136, opened 2026-08-23
Corpus nav_junk check fires on our own contents heading and misses nine real duplicate listings


jloutsch:
Split out of #132. The detector that was supposed to catch duplicate contents listings is wrong in both directions.

## False positive

`test_corpus_book_invariants` asserts `nav_junk == 0`, counting body blocks whose text is one of four literal strings:

```python
NAV_MARKERS = frozenset(
    {"Page List", "Navigation", "Begin Reading", "Table of Contents"}
)
```

pg64317 is the one book failing it. It fails on kfxgen's **own generated contents page**. The generator emits `chapter["title"]` as the heading chunk (`native_generator.py:2810`); that book's TOC label is the literal string "Table of Contents", which is in the marker set. The book's actual inline toc was recognised and discarded correctly.

Every other corpus book has a TOC label of "Contents" or "CONTENTS", which is why only this one trips it.

Related: `_rebuild_contents_page` sets `contents_ch["text"] = "Contents\n\n" + entries`, intending "Contents" as the header. The generator ignores `text` entirely once `toc_links` is set, so that normalisation never runs. Either the generator should use it, or it should go — but the assertion should not depend on which label a source book happened to use for its TOC entry.

## False negatives

Eight corpus books ship a visible duplicate contents listing. `nav_junk` catches one of them, and that one is the false positive above. The other seven leak silently and the suite passes.

They are invisible to the check because their listing heading is "CONTENTS" or "CONTENTS." — neither is in `NAV_MARKERS`, and adding them would not generalise. A book whose listing has no heading at all would still pass.

The marker set was written for #60, where the leak genuinely was a `page-list` / `landmarks` nav document with a fixed label. It was never a duplicate-listing detector; it is being read as one.

## What a real check looks like

Measure the source against the output rather than pattern-matching output text. Method that produced the count above:

1. Collect every `<a>` label inside a `class~="toc"` element in the source EPUB.
2. Convert, and for each output chapter take its block texts in order.
3. Find the longest **contiguous** run of blocks that are all source toc labels.
4. A long run means the listing survived as a listing.

Contiguity is the part that matters. A set-membership test counts "CHAPTER I" as leaked when it is just the real chapter's heading — that inflated a first pass from 8 books to 10, including two that have no leak at all.

Results, for whoever writes the assertion:

```
pg1998   toc_entries= 62  longest_run= 62   <- leak
pg2701   toc_entries=138  longest_run=138   <- leak
[...]

PR #138, opened 2026-08-23
fix: never title a chapter with an image token (#133)

### LINKED ISSUE #133: A cover image token becomes a chapter title, so every generated contents page lists an image as an entry
Found while measuring #132. Independent of it, and larger.

## Symptom

Every generated contents page in the corpus lists the cover image as its first entry. The entry's label is a raw IMG token, so the cover is re-rendered inside the table of contents, as a link.

From the KFX for pg64317:

```
'Table of Contents'
<IMG token — the cover image>
<title/author line>
'I' 'II' 'III' 'IV' 'V' 'VI' 'VII' 'VIII' 'IX'
```

The second junk entry is the title-page chapter, which is a smaller problem but has the same origin.

## Scope

**39 of 39** corpus books that build a contents page. Not a subset — all of them.

```
books with a rebuilt contents chapter: 39; with IMG-token entries: 39
```

## Cause

`_leading_chapter_title` (`plugin/kfxgen/converter.py:789`) titles the leading front-matter chapter from its first block:

```python
if head_blocks:
    t = (head_blocks[0].get("text") or "").strip()
    if 0 < len(t) <= _LEADING_TITLE_MAX_LEN and "\n" not in t:
        return t
return "Front Matter"
```

An IMG token is short and contains no newline, so it passes both guards. The chapter is then titled with an image.

That title propagates. `_rebuild_contents_page` skips chapters whose title is in `CONTENTS_SKIP_TITLES` — a set of literal strings ("cover", "contents", "title page", …). An IMG token matches none of them, so the chapter becomes a `toc_links` entry, and the generator emits `link["text"]` as a chunk (`native_generator.py:2811`). The token is processed into an image reference downstream, the same as any body image.

The guard is length-and-newline only. It has no notion of the text being a token rather than words.

## Fix

Reject a candidate title that is, or contains, an image token, and fall back to `"Front Matter"`. That fixes the label at its source, so nothing downstream needs to know about it — `CONTENTS_SKIP_TITLES` keeps matching on literal strings, and the heading-emission path in the generator is untouched.

Worth checking at the same time whether the *rendered* chapter heading for these front-matter chapters also carries the token, not just the contents entry. `body_is_image_only` in `native_generator.py:2796` suppresses a heading for image-only chapters, so it may already be covered, but that path keys on the body rather than the title.

Second junk entry (title-page chapter) is the same class of problem — a chapter whose title is not a recognised label — but not the same cause, and probably wants its own decision about whether a title page belongs in a generated contents listing at all.

## Verification

Regression test should build a contents page from chapters where the leading one starts with an image, and assert no `toc_links` entry contains an image token. Mutate `_leading_chapter_title` to confirm the assertion discriminates rather than passing vacuously.

Device check wanted for how the duplicated cover reads on the contents page, per #123.

jloutsch:
Closes #133.

## The bug

`_leading_chapter_title` took the first block's text as the chapter title whenever it was 1–60 chars with no newline:

```python
if 0 < len(t) <= _LEADING_TITLE_MAX_LEN and "\n" not in t:
    return t
```

`_make_img_token` emits a single run with its spaces escaped, so a bare cover image passes both guards. The leading front-matter chapter was titled with a picture.

That title reached the reader twice. Once as the chapter's heading chunk, and once as a contents entry — `_rebuild_contents_page` skips chapters by matching literal strings, so nothing in `CONTENTS_SKIP_TITLES` could match a token.

## Worse than the issue described

Neither copy rendered as an image. The generator resolves tokens in body text, not in titles, so both were emitted as **control-character strings printed to the reader**.

Corpus-wide that is **116 unresolved tokens across all 77 books** — not only the 39 with a generated contents page. #133 was measured through the contents path and undercounted itself.

## The fix

One condition: reject a candidate title containing an image token, fall back to `Front Matter`.

Rejecting rather than stripping is deliberate. What survives stripping is alt text or a caption fragment, which names the picture rather than the section.

## Verification

Before/after over all 77 books, comparing full conversions:

| | before | after |
|---|---|---|
| raw image tokens in emitted text | 116 | **0** |
| books affected | 77 | **0** |
| image resource count | — | unchanged in every book |
| block count | — | unchanged in every book |

No image is lost — the cover still comes from the synthetic cover chapter (`_is_cover`, #32). The only text delta is −2899 chars, exactly the tokens replaced by shorter labels; no book grew.

Tests, each watched failing first:

- `_leading_chapter_title` rejects a bare token, and rejects a caption carrying one
- a parametrized guard that the existing length/empty/newline rules still hold
- the assembly path emits `Front Matter` and keeps the image in the body
- the reader-visible symptom: no contents entry may carry a token
- a new corpus invariant, `raw_img_tokens == 0`

Gates: 747 unit+integration pass, 13 tier3_strict pass, ruff check and format clean. Corpus sweep 76/77.

## Known remaining failure

pg64317 still fails `nav_junk` — kfxgen's own generated `Table of Contents` heading matching `NAV_MARKERS`. That is #136 and fails identically on main. Not touched here.

## Not addressed

The front-matter chapter now prints a visible `Front Matter` heading where it previously printed a control-character string. Readable, and consistent with what books already got when the first block was long or multi-line — but whether that page should carry a heading at all is a separate call, so I left it.

### DIFF
--- a/plugin/kfxgen/converter.py
+++ b/plugin/kfxgen/converter.py
@@ -790,10 +790,27 @@ def _leading_chapter_title(head_blocks):
     """Title for front matter that precedes the first TOC anchor.
 
     Use the first block's text when it is short enough to be a heading,
-    otherwise a neutral 'Front Matter' label."""
+    otherwise a neutral 'Front Matter' label.
+
+    An image token is rejected however short it is. The length-and-newline
+    guards alone let one through — `_make_img_token` emits a single run with
+    its spaces escaped, so a bare cover image reads as a brief, tidy heading.
+    The chapter was then *titled* with a picture, and because
+    `_rebuild_contents_page` skips chapters by matching literal strings, no
+    entry in `CONTENTS_SKIP_TITLES` could match it: the cover came back as a
+    contents entry, rendering the image inside the listing. Every book in the
+    corpus that builds a contents page had one. (#133)
+
+    Rejecting rather than stripping the token is deliberate. What remains after
+    stripping is alt text or a caption fragment, which names the picture rather
+    than the section — 'Front Matter' is the more honest label."""
     if head_blocks:
         t = (head_blocks[0].get("text") or "").strip()
-        if 0 < len(t) <= _LEADING_TITLE_MAX_LEN and "\n" not in t:
+        if (
+            0 < len(t) <= _LEADING_TITLE_MAX_LEN
+            and "\n" not in t
+            and not _IMG_TOKEN_RE.search(t)
+        ):
             return t
     return "Front Matter"
 
--- a/plugin/kfxgen/converter.py
+++ b/plugin/kfxgen/converter.py
@@ -792,25 +792,24 @@ def _leading_chapter_title(head_blocks):
     Use the first block's text when it is short enough to be a heading,
     otherwise a neutral 'Front Matter' label.
 
-    An image token is rejected however short it is. The length-and-newline
-    guards alone let one through — `_make_img_token` emits a single run with
-    its spaces escaped, so a bare cover image reads as a brief, tidy heading.
-    The chapter was then *titled* with a picture, and because
-    `_rebuild_contents_page` skips chapters by matching literal strings, no
-    entry in `CONTENTS_SKIP_TITLES` could match it: the cover came back as a
-    contents entry, rendering the image inside the listing. Every book in the
-    corpus that builds a contents page had one. (#133)
+    Image tokens are stripped before the guards run. The guards alone let one
+    through — `_make_img_token` emits a single run with its spaces escaped, so
+    a bare cover image reads as a brief, tidy heading. The chapter was then
+    *titled* with a picture, and because `_rebuild_contents_page` skips
+    chapters by matching literal strings, no entry in `CONTENTS_SKIP_TITLES`
+    could match it: the cover came back as a contents entry. Worse, a title is
+    emitted as a heading chunk without token resolution, so both copies reached
+    the reader as raw control characters — 116 of them across the corpus. (#133)
 
-    Rejecting rather than stripping the token is deliberate. What remains after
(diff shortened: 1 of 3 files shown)


jloutsch:
Device check on this fix, and one gap in its gate.

## The fix is verified on hardware, and #133 is worse than written

Paperwhite 11th gen (5.19.2), one corpus book (pg64317) converted twice from identical source — once by `main`, once by a branch carrying this change.

The `main` build **crashed the reader** while paging through its opening pages. The build with this fix paged the same book from the opening through to chapter IX and the licence page with no crash. Full note on #133.

The unresolved token is not only a contents entry. It also lands in `$145` `content_1` at character 0 — the first content fragment of the book, which is the region that crashed — as raw `\x00`/`\x01` bytes.

## The gate here has a blind spot

```python
"raw_img_tokens": sum(1 for t in texts if IMG_TOKEN_RE.search(t)),
```

`texts` is built from `$145` chunks, so this sees the body copies but not the navigation ones. The crashing book carried 4 tokens: 2 in `$145`, and **2 in `$389`**, as the first and last entries of the Go To list.

They diverge because a chapter title reaches the body as a heading chunk only when `_omit_title_heading` is unset, but it reaches `$389` either way. A title-borne token on a heading-suppressed chapter would therefore pass this assertion while still shipping control bytes to the device — which is the exact path that produced the crash.

Not a criticism of the check: the source-level fix here removes both, and asserting on emitted text is the right instinct. It is that the assertion is now gating a crash rather than a cosmetic defect, so the blind spot matters more than it did when written.

#137 adds a complementary whole-container scan (`2050af8`) covering nav labels, section names and metadata. Verified discriminating against real builds: `main`-built containers score 4 and 3, fixed builds 0 and 0. If these land in either order the two checks coexist — they assert different things and neither replaces the other.

Worth landing this first; #137 adopts this PR's `_leading_chapter_title` change verbatim so the two agree rather than conflict.

PR #144, opened 2026-08-31
fix: stop inventing navigation entries the source never had (#143)

### LINKED ISSUE #143: Navigation pane carries invented and machine-generated labels: 'Front Matter' in 77/77 books, raw filenames in 14
Reported from reading: navigation panes carry entries like "Front Matter", "back matter", `bm_001` — structural or machine labels for pages that are publisher back matter ("stay in touch with the author", a note to the reader), not chapters anyone navigates to.

kfxgen produces both shapes. Measured over the 77-book corpus by building each book's chapter list and taking the labels that reach `$389` (i.e. excluding `_omit_from_toc`):

```
6,231 nav entries across 77 books
  "Front Matter"      77 entries in 77 books
  filename-titled  1,213 entries in 14 books
```

Two different causes, same symptom and same place to fix.

## 1. "Front Matter" — every book, and the two lists disagree

`_leading_chapter_title` (`converter.py:908`) labels front matter that precedes the first TOC anchor with the invented string `LEADING_TITLE_FALLBACK` (`converter.py:879`). No book contains those words; kfxgen writes them.

The inconsistency is the tell. That label **is** already excluded from the contents page kfxgen generates — it is in `CONTENTS_SKIP_TITLES` (`converter.py:1330`) — so the code already holds the view that a reader should not see it. The nav pane is built from `chapter["title"]` filtered only by `_omit_from_toc` (`native_generator.py:2323`), which is set on the synthetic cover chapter and nothing else. So the same label is suppressed in one list and printed in the other.

**Provenance, since it is recent and mine.** Before #133 this entry existed in all 77 books too, as a raw image token — `\x00IMG\x01...` control characters, the ones device-verified as crashing a Paperwhite. #133 replaced them with `Front Matter`:

```
                     before #133          now
"Front Matter"       0 in  0 books    77 in 77 books
raw IMG token       77 in 77 books     0 in  0 books
```

So this is not a regression — it is a strictly better label than what preceded it. It is an invented one left in front of the reader, which the earlier garbage disguised.

## 2. Filename-titled entries — 1,213 across 14 books

Tail orphans (spine items after the last TOC coordinate) are titled from the filename (`converter.py:1053`):

```python
stem = norm.rsplit(".", 1)[0] if "." in norm else norm
ch = _flag_nav(_mk(stem or f"Section {si + 1}", spine_blocks[si]), ...)
```

which on Gutenberg's image-wrapper pages yields entries like

```
1430017453568371517_0028.jpg.id-5936110575048331346.wrap-0.html
```

85 characters of machine string in a Go To menu.

| book | nav entries | filename-titled | longest |
|---|---|---|---|
| pg22210 | 994 | **835 (84%)** | 64 |
| pg12082 | 244 | 237 | 85 |
| pg120 | 89 | 43 | 64 |
| pg394 | 47 | 26 | 64 |
| pg98 | 68 | 16 | 63 |
| pg1400 | 80 | 15 | 63 |
| pg1260 | 57 | 14 | 64 |
| pg25851 | 234 | 11 | 82 |
| pg28428 | 148 | 10 | 71 |

Nine more books have 1–2 each. **Pre-existing and unchanged** by the recent work — 1,213 before #133 and 1,213 now.

These pages reach the tail-orphan path rather than being skipped as image-only because they carry a `back` link, so `_has_real_text` is satisfied by one word.

## Why no existing check sees it

Every tier reasons about structure, not about whether a label is meaningful. Tier 1 checks invariants, tier 2 decodes, tier 3 diffs against our own goldens, and the corpus sweep counts blocks, anchors and images. A nav entry titled `..._0028.jpg.id-...wrap-0.html` is structurally perfect. This is the same gap as #133: correct bytes, wrong thing in front of the reader, and only visible by looking.

## Options

1. **Omit rather than rename.** `_omit_from_toc` already exists for the synthetic cover; setting it on the synthetic-label chapter and on filename-titled orphans is the smallest change and matches what `CONTENTS_SKIP_TITLES` already does for the contents page. Removes reachable entries, which is the part worth checking on hardware.
2. **Title from content.** Use the page's first heading or first words instead of the filename. Better for a back-matter page with a real heading ("A Note from the Author"), useless for an image wrapper whose only text is "back".
3. **Neutral label.** "Illustration", "Section N". Honest, still not something a reader wants 835 of.
4. **Do nothing for (1), fix (2).** Defensible: "Front Matter" is at least words, while an 85-character filename is not.

(1) and (2) combine well — title from content where there is content, omit where there is not.

## Priority

Medium. Nothing is broken and no text is lost; the navigation pane is just full of things that are not chapters, in 14 of 77 books severely. It is user-visible on every book (the "Front Matter" row) and disfiguring on illustrated ones.

Wants a device check before merging, per #109 — the `toc_navigation` checklist entry — because omitting entries changes what the nav pane offers, and the pane is the project's headline feature.

jloutsch:
Closes #143.

Implements the rule you arrived at from reviewing source EPUBs: the publisher's TOC ends at the last real section ("End Notes", "Index"), and the back matter after it is deliberately unlisted — so kfxgen should not invent entries for it.

## The rule is mechanical, not a matter of taste

Both paths that produce an invented label exist **because** the source TOC never referenced those spine items — that is literally how they are defined. `_leading_chapter_title` labels content before the first TOC anchor; tail orphans sit past the last one.

Verified: the chapters carrying invented labels are exactly the chapters whose title is not a source TOC label.

| book | not a source TOC label | = |
|---|---|---|
| pg22210 | 836 | 1 `Front Matter` + 835 filenames |
| pg12082 | 238 | 1 + 237 |
| pg120 | 44 | 1 + 43 |
| pg64317 | 1 | 1 + 0 |

## Effect

```
before  nav entries=6231  "Front Matter"=77  filename-titled=1213
after   nav entries=4941  "Front Matter"= 0  filename-titled=   0
```

1,290 entries removed — exactly 77 + 1,213. No other label moved.

pg22210 goes from 835 of its 994 nav entries being 64-character filenames to none.

## Narrower than the issue proposed, on purpose

I did **not** omit every untoced chapter. A head chapter whose first block genuinely reads as a heading ("Copyright 2026") keeps its entry: that label is the book's own words rather than ours, and dropping it would remove a usable destination in order to fix a naming problem.

The corpus outcome is identical either way — all 77 heads there open with a cover image, so they have no title of their own — which means the narrower rule costs nothing measurable and risks less. Both behaviours are pinned by tests.

## Omitting from the TOC is not dropping content

`_omit_from_toc` is read only where the nav pane is built (`native_generator.py:2323`). These pages still ship in the reading flow and still page through. There is a test for that specifically, because it is the property that makes this safe.

There is also a guard test: a book whose TOC resolves to nothing has *every* chapter untoced, and omitting them all would leave a book with no navigation whatsoever — worse than a machine label. That path is untouched.

## Verification

- suite **840 passed**, tier3_strict 13 passed, corpus sweep 154 passed
- `publisher_structure`'s golden moves: its `add_manifest_item(in_spine=True)` entries are spine-only and so read as tail orphans. That absence from the NCX is an artifact of the fixture builder rather than modelled publisher behaviour; the fixture's pins are on link resolution and text, and all still pass. Every other golden is byte-identical.

## Before merging

Wants a device check, per #109's `toc_navigation` entry. This changes what the navigation pane offers, and the pane is the project's headline feature — the thing to confirm is that the entries that remain still land correctly and that nothing a reader wants has gone missing.


jloutsch:
Device-check pair built, and building it turned up a side effect worth pinning before this merges.

## The opening landmark moves

The start-of-content landmark (`$236` container, `$238: $233`) is built from the first *listed* nav entry, so omitting the front-matter chapter also moves where the book opens:

| | landmark points at |
|---|---|
| before | `Front Matter` — the Gutenberg boilerplate page |
| after | the title page |

That is arguably the better target, but it was a side effect of the nav change rather than a decision, and nothing pinned it. Two tests now do (`13b330d`), so a future change to `_omit_from_toc` cannot move the opening position silently.

It is the first thing the device check asks about.

## Files

`~/Desktop/kfxgen-devicecheck-143/`, two pairs from real corpus books, distinct titles and ASINs:

**pg64317, precision** (1.2 MB) — exactly one entry differs, 13 -> 12. Short enough to read the whole list by eye and confirm nothing else moved. Also where the landmark shift is visible: open each fresh and note which page it lands on.

**pg22210, scale** (103 MB) — 991 -> 155 nav entries, 835 raw filenames -> 0. The question there is not whether the junk went but whether anything wanted went with it.

The README states plainly that the removal claim — "these were never in the source TOC, so nothing a reader wants is lost" — is what is under test rather than a given, and asks for the shape of any entry that should have stayed.

Suite 842 passed.

jloutsch:
## Device check: passed on a Paperwhite

| Device | Firmware | Status | Check | Result |
|---|---|---|---|---|
| Paperwhite 11th gen (2021) | 5.19.2 | current | Navigation pane entries land on their chapter | **pass** |

Two books, each converted from identical source by `main` and by this branch.

**pg64317 — precision.** 13 -> 12 nav entries. `Front Matter` gone, everything else present, and the remaining entries still land on their chapters. That is the over-removal check: the intended entry went and nothing else moved.

**pg22210 — scale.** 991 -> 155 nav entries, 835 raw filenames -> 0. Both builds carry **995 sections**, so omitting 836 nav entries dropped no content — which was the property most worth confirming, since the whole change is about removing navigation.

## The landmark side effect is inert

Flagged above: omitting the first chapter moves the start-of-content landmark, because it is built from the first *listed* nav entry.

| | landmark | position |
|---|---|---|
| before | `Front Matter` | 1006 |
| after | `IN THE FORBIDDEN LAND` | 1028 |

It moved, and to a better target — the book title rather than Gutenberg boilerplate. But **both builds open at the cover**, confirmed on device for the before and after build independently. The Paperwhite opens a freshly sideloaded book at the start of the reading order and does not consult the landmark for first-open, so the shift changes nothing a reader experiences on this firmware.

Worth stating what that does *not* establish: it was checked on 5.19.2 only. The Oasis on terminal 5.18.2 was not run, and a reader who opens via a stored position rather than a fresh sideload was not exercised. The landmark is now more accurate either way, so the risk of the shift is low, but "inert" is a claim about one device and one path.

The two tests added in `13b330d` pin the coupling regardless, so a future change cannot move the opening position without saying so.

Sign-off recorded via `scripts/device_signoff.py`, per #109.

Issue #112, opened 2026-08-21
No oracle compares kfxgen's fragments against Amazon's own output


jloutsch:
Prompted by [HankunYu/kindle-comic-workaround-5.19.x](https://github.com/HankunYu/kindle-comic-workaround-5.19.x), an independent comic→KFX writer that verifies itself by diffing its container **fragment by fragment** against the Kindle Create → KFX Output toolchain. kfxgen has no equivalent, and the gap is worth naming precisely because it is smaller than it first looks.

## What kfxgen already compares

| Oracle | Compares | Against |
|---|---|---|
| tier-1 | in-process invariants | itself |
| tier-2 | fragment *types* | upstream `kfxlib`'s decode of **our own** output |
| tier-3 | fragment-type counts, per-type top-level key sets | **our own** committed goldens |
| tier-3 strict | SHA-256 of the file | **our own** committed goldens |
| `research/kfx-format-baseline/` | fragment types + symbol sets | **Amazon's** output, but only against a prior snapshot of Amazon's output |

Every layer that inspects structure compares kfxgen to kfxgen. The one layer that touches Amazon's output tracks *its* drift over time; it never asks whether kfxgen's fragments look like Amazon's.

So kfxgen can prove it is self-consistent, reproducible, and decodable by an independent parser. It cannot currently answer: **where kfxgen and Amazon make the same choice, does kfxgen use the same fragment shape?**

## Not the same as byte-equivalence

Worth stating up front, because the source of the idea aims at byte-equivalence and kfxgen should not.

HankunYu is *replicating* the Kindle Create → KFX Output toolchain for fixed-layout comics. Byte-equivalence is their specification, and it is tractable because comics are images plus minimal structure.

kfxgen is a different generator making deliberate different choices for arbitrary reflowable EPUBs — text chunking, position maps, anchors, embedded fonts, inline emphasis. Its output *should* differ from Amazon's. The goal here is **conformance**, not equivalence: same idiom where the choice is the same.

## How far the existing machinery already gets

`research/kfx-format-baseline/kfx_inventory.py` decodes any `.kpf`/`.kfx` through the full upstream `kfxlib` and emits fragment types + symbol sets, with a `--diff` mode. It already reads both sides. What it captures is inventory-level; what a conformance check needs is per-fragment-type **top-level key sets** — which `test_golden_corpus.py` already computes, just against our own goldens.

[...]

Issue #79, opened 2026-08-01
dangling == 0 is a vacuous invariant — dropped links are invisible to it


jloutsch:
Found while investigating #76.

kfxgen **drops** a link whose target cannot be resolved rather than emitting a dangling `$179`. So `set(targets) - anchors` stays empty whether links resolve correctly or are never emitted at all. The invariant is satisfied perfectly by a build that emits zero links.

## Where this bites

- `tests/integration/test_public_corpus.py` — the corpus invariant, which runs weekly in CI and is the gate specifically intended to catch silent damage
- `tests/integration/test_golden_corpus.py` — the `publisher_structure` shape assertion

A regression that stopped emitting links would pass both. The golden fixture has partial cover, since its structural fingerprint compares `$266`/`$179` key sets and would notice counts changing — the corpus sweep has no such backstop.

This is the same shape of mistake as the golden fingerprint tolerating text changes (which let a 44% text-loss bug through): an assertion that looks like it constrains something but does not.

## Fix

Assert an expected link **count**, not just the absence of dangling targets:

- corpus invariant mode: assert links > 0 for books known to contain internal links, or record and compare counts
- golden shape test: assert the exact number of link spans the fixture should produce

The baseline-diff mode of the corpus test already compares `links` per book and is not affected — the gap is in the invariant mode, which is the part that runs unattended.

jloutsch:
Fixed in #75.

The golden fixture now asserts an exact link count (`EXPECTED_PUBLISHER_LINKS`), and the corpus invariant asserts that a book emitting anchors also emits spans for them.

Verified by simulating the regression — forcing every link to drop — rather than by inspection:

```
golden shape test:  2 failed
corpus invariants:  1 failed (on a real Gutenberg book)
```

Before the change both passed, which was the whole problem.

PR #146, opened 2026-09-02
fix: an image is not a heading — drop $790 from image entries (#120)

jloutsch:
Split out of the #120 spike, which is otherwise discarded. Refs #120.

This part stands alone: no token change, no conformance for its own sake, just a flag kfxgen sets that it should not.

## What it does

`$790` (`-kfx-heading-level`) marks a navigable section start. kfxgen applied it to entry 0 of every storyline whatever that entry was, so a chapter opening with an image had its **picture flagged as a heading**. Across eight corpus books, 19 image entries carry it.

## Why it needed checking, not asserting

I nearly called this a safe one-liner. It isn't obviously safe.

`native_generator.py:3002` records that a TOC target must be a leaf with a `$145` content reference and *ideally* a `$790:1` marker, or Kindle treats the position as non-navigable. An image entry has no `$145` either — so on that reading, `$790` might have been the only thing keeping image-first chapters tappable, and removing it would be the #51/#62 failure all over again.

**Amazon settles it by counterexample, at scale.** Decoding Kindle Previewer 3.106 output:

| | |
|---|---|
| storylines that *begin* with an image (3 books) | **1,084** |
| of those, carrying `$790` | **0** |
| image entries total (4 books) | 1,698 |
| of those, carrying `$790` | **0** |

Whatever makes such a section navigable, it is not this flag.

## Tests

Two, and the second is the one that matters:

- no image entry is flagged as a heading
- **a chapter whose first entry is text still gets its marker** — a section with no `$790` anywhere is a TOC entry that does nothing on tap, and that is the regression this could have introduced

Reproducing the bug needs `_omit_title_heading` on the chapter. Without it the title is emitted first and takes the flag, so the obvious fixture passes vacuously — my first version did.

## Verification

- suite **844 passed**, tier3_strict 13 passed
- two goldens regenerated: `with_cover` and `unused_manifest_image`, both of which open on an image. Every other golden byte-identical.

## Before release

Worth a device check on a book with image-first chapters — pg120 has seven — per #109's `toc_navigation` entry. The risk being retired is theoretical and the evidence against it is Amazon's rather than ours, so confirming a TOC tap still lands is cheap insurance.


jloutsch:
## Closing: this is a regression, and my justification was a false comparison

Asked whether this needed a device test. It needed a structural one first, and that test fails.

### The change breaks TOC entries

`native_generator.py:3002` says a nav position must resolve to a leaf with `$145` or `$790`, or Kindle treats it as non-navigable and the tap is a no-op. Measuring nav targets before and after:

| book | nav entries landing on a position with neither, on `main` | on this branch |
|---|---|---|
| pg120 | 0 | **6** |
| pg1184 | 0 | **4** |
| pg245 | 0 | 0 |
| pg40739 | 0 | 0 |

Ten TOC entries across two of four books become no-ops. That is the #51/#62 failure exactly.

### Why the Amazon evidence did not mean what I said

I argued: Amazon has 1,084 storylines beginning with an image and none carry `$790`, so the flag is not what makes such a section navigable.

True, and beside the point. **Amazon’s nav never points at an image entry** — 0 of 25 nav entries across pg12082, pg22210 and pg40739 land on one. It can omit the flag because it never navigates there.

kfxgen does navigate there: it points each nav entry at the chapter’s first chunk whatever that chunk is. So `$790` on an image entry is not a stray heading flag — it is load-bearing for kfxgen’s navigation model, and the `if i == 0` that sets it is the same condition that makes that entry a nav target. There is no safe subset to remove.

The two pipelines differ in design, not in correctness, and this is conformance-for-its-own-sake of the kind #112 warns about: *a difference here is not a defect*.

### What would be needed to do it properly

Point the nav at a navigable entry the way Amazon does, then drop the flag. That is a change to chapter-start position selection — the machinery behind #51, #62 and #79 — for no observable gain, on the same reasoning that closed #120. Not worth it.

Closing unmerged. The fixture and the two tests were the useful part: reproducing the shape needs `_omit_title_heading`, and the control that a text-first chapter keeps its marker is what caught this.

Issue #96, opened 2026-08-07
Output is not byte-reproducible: image style symbols allocated from an unordered set


jloutsch:
Converting the same EPUB twice produces two different KFX files. Found while verifying #79; unrelated to it and pre-existing.

## Impact

Reading is unaffected — every `$145`, `$259`, `$265` and every `$266` anchor is byte-identical across runs (checked on a book with 2,367 anchors and 3,948 content positions). The variance is confined to three fragments:

| fragment | how it differs |
|---|---|
| `$ion_symbol_table` | same 2,832 symbols, different **order** — first divergence at index 2734, `s_img_page` vs `s_img` |
| `$270` `$181` | same 2,620 entries as a multiset; symbol ids shift, e.g. `[157, 3587]` vs `[157, 3586]` |
| `$419` `$252` | content-fragment name list ordering |

The last two are consequences of the first: once local symbols are created in a different order their ids shift, and everything referencing them by id shifts with them.

The practical cost is to the corpus A/B diff (see `research/baseline_runner.py`). Two builds of the same book report a difference in these three fragments even when nothing changed, so real churn has to be picked out of standing noise. It also means "byte-identical" can't be used as a clean regression signal for image-bearing books — the golden fixtures are stable only because they use a single minimal JPEG.

## Cause

`native_generator.py:3068`:

```python
kinds_used = set()
for c in all_chunks:
    if ... == "image":
        kinds_used.add(_classify(w, h))

for kind in kinds_used:            # <-- iteration order of a set of strings
    name = {"small": "s_img_sm", "inline": "s_img", "page": "s_img_page"}[kind]
    ...
    extra_style_names.append(name)
```

Iterating a set of strings is not stable across processes — CPython randomizes string hashing per interpreter unless `PYTHONHASHSEED` is fixed. A book using more than one image size class allocates those `$157` image styles in whatever order the set yields, and every local symbol created afterwards is numbered from there.

This only shows up when at least two of `small` / `inline` / `page` are present, which is why single-image fixtures never caught it.

## Fix

Iterate in a fixed order rather than the set's, e.g. over the classification names in a declared sequence and skip the ones not used. `referenced_targets` at line 2911 already does the equivalent with `sorted(...)`; this site was missed.

## Verification

A test that converts the same input twice in separate processes and asserts the outputs are byte-identical would pin it — same-process runs may not reproduce it, since hash randomization is per interpreter.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #147, opened 2026-09-02
fix: size images from their own width, not a fixed 9.626% (#145)

### LINKED ISSUE #145: Images narrower than 600px render at 9.626% width — 55% of corpus images, including 10 covers
Found on device while checking #120. Reported first as "the image strip is very small and I cannot tap it to enlarge — and the cover is very small too."

## What happens

Every image that is not classified `page` or `small` gets `$56: 9.626%`, and `$56` is **width**. So it renders at under a tenth of the text width.

`native_generator.py:3202` classifies by pixel size:

```python
if w >= 600 and h >= 600:   return "page"     # $56 = 100%
if w <= 300 and h <= 300 and ratio <= 1.4:
                            return "small"    # $56 = 3em
                            return "inline"   # $56 = 9.626%
```

Anything narrower than 600px falls through to `inline`. That is not a rare corner:

```
corpus images        4,436
  page      1,625     36.6%   ($56 = 100%)
  inline    2,462     55.5%   ($56 = 9.626%)   <- renders tiny
  small       349      7.9%   ($56 = 3em)

books with at least one 9.626% image : 29 of 77
covers landing in that class         : 10
worst: pg22210 (339) · pg24855 (305) · pg1184 (302) · pg245 (291) · pg28428 (284)
```

**Ten covers** are affected, which is the "cover is very small" half of the report — a 462x700 cover is under 600 wide, so it is not `page`.

## What Amazon does instead

Decoded Kindle Previewer 3.106's own output for the same book (pg40739, whose images are 512x206):

| | `$56` (width) | style |
|---|---|---|
| Amazon, ordinary image | **100** | one style per image |
| Amazon, four images sharing a container | 32 | per-image, with `$57` height varying 12.3–12.9 by aspect |
| kfxgen, all of them | **9.626** | one `s_img` style for the whole book |

Amazon sizes each image individually and defaults to full width. kfxgen has one style per size-class per book and defaults to 9.626%.

## Where 9.626 came from

It is not arbitrary, which is why this survived. The comment at `native_generator.py:1378` records it as observed in a jhowell KFX Output reference file for *"wide rule-style decorations (e.g. 1090x92)"* — a thin horizontal rule, where a small height fraction is right.

The defect is the fallback: a value correct for a 1090x92 rule became the default for every image between 300px and 600px, which is most illustrations in scanned books. The comment also describes `$56` as capping *height*; Amazon's usage — 100 for a full-width image, 32 for four across — reads as width.

## Not the same as tap-to-zoom

The report also notes the image cannot be tapped to enlarge. That is separate and probably not a defect: #118 established that Amazon does not emit zoomable plugin resources either. Worth keeping apart from the sizing question.

## Suggested

Default to `$56: 100%` and let `$65` (max-width) and the aspect ratio do the work, keeping `small` for genuine ornaments. That matches what Amazon does for everything except deliberately grouped images.

Wants a device check before merging — this is a pure rendering change, invisible to every structural tier, and the only way to know it looks right is to look. The `body_images_render` entry in the tier-4 checklist covers it.

## Priority

High for a reader. It is not data loss and nothing is broken structurally, but it makes illustrated books look wrong on more than half the images in the corpus, and it is the first thing anyone notices — which is how it was found.

jloutsch:
Closes #145.

## The defect

Every image not classified `page` or `small` got `$56: 9.626%`. **`$56` is width** — confirmed in upstream kfxlib's property table (`yj_to_epub_properties.py:545`), where our own comment calling it a height cap is simply wrong. Those images rendered at under a tenth of the text column.

```
corpus images        4,436
  page      1,625     36.6%   width 100%
  inline    2,462     55.5%   width 9.626%   <- slivers
  small       349      7.9%   width 3em

books affected: 29 of 77.  Covers in that bucket: 10.
```

Not an invented number: copied from a jhowell reference for a 1090x92 horizontal rule, where a small fraction is right. It became the default for every illustration between 300 and 600 pixels.

## The rule, recovered rather than guessed

Dividing each sub-100% width in decoded Kindle Previewer output by the image's pixel width yields one constant — a **496px text column**:

```
width% = min(100, pixel_width / 496 * 100)
```

**1,411 of 1,422** percent widths across seven books (99.2%). The residue: 7 entries against a 480px reference, one against 729, and 3 small images given 100% anyway.

An image at least as wide as the column fills it; a narrower one keeps its natural proportion instead of being shrunk to a sliver. The old `page` bucket is subsumed as the capped case. `small` stays — a square dingbat is not an illustration, and stretching it to 45% of the column would be a new bug.

## Agreement with Amazon after the change

| book | kfxgen %-sized | amazon %-sized | identical widths |
|---|---|---|---|
| pg19771 | 18 | 18 | **18** |
| pg22210 | 1,072 | 1,086 | **1,039** |
| pg65580 | 96 | 67 | 65 |
| pg40739 | 8 | 3 | 3 |

The remaining gap is where Amazon uses `em` for grouped or linked images — the inline-render distinction closed in #120 as visually indistinguishable.

## Reproducibility

Styles are now allocated per distinct width rather than three per book (119 for the largest corpus book, and Amazon uses per-image styles too). Allocated in sorted order, so ordering is a property of the widths and not of set iteration — the #96 rule. Verified: two conversions of one book are byte-identical.

## Verification

- suite **849 passed**, tier3_strict 13 passed
- four goldens regenerated, all image-bearing; the rest byte-identical
- one test asserts ornaments keep `3em`, because the obvious over-reach here is scaling dingbats to full width

## Device check

Files staged at `~/Desktop/kfxgen-devicecheck-145/`. Unlike #120 — where identical output was the honest answer — **a visible difference here is the point**. pg120 is the careful case: 80 slivers should grow, 44 already-correct images should not move, and 75 ornaments must stay small.

Not merging until that comes back.


jloutsch:
## Device check: passed on a Paperwhite

| Device | Firmware | Status | Check | Result |
|---|---|---|---|---|
| Paperwhite 11th gen (2021) | 5.19.2 | current | Every image the source displays is drawn | **pass** |

pg40739, same book and same two landmarks as the #120 check. In the `main` build the images render small; in this build they are much larger and easily viewable, and the cover fills the column rather than appearing as a sliver.

That is the outcome that was in doubt. #120 closed because a measured difference turned out to be invisible; this one is visible, which is what makes it worth shipping.

## What was not device-checked, and why that is acceptable

The ornament case — that 75 small dingbats in pg120 keep their `3em` sizing rather than ballooning to full width. It was not separately confirmed on hardware.

It is unchanged **by construction** rather than by luck:

* the `small` classifier and its `3em` style are untouched by this change
* an image cannot be both `>= 600px` square (old `page`) and `<= 300px` square (`small`), so reordering the checks is inert
* measured in the built files: 75 ornaments at `3em` before, 75 at `3em` after

A unit test pins it as well. So there is no before/after difference for an ornament to expose.

Sign-off recorded via `scripts/device_signoff.py`, per #109.

Issue #33, opened 2026-07-01
Verify #9 block-level CSS actually applies in real Calibre conversions (Stylizer opts silent-failure found in #15)


jloutsch:
## Summary

While verifying #15 with a real `ebook-convert` run, the conversion log showed:

```
Stylizer unavailable ('OEBBook' object has no attribute 'opts'); skipping per-element CSS
```

The per-element style resolver (`converter.py::_build_style_resolver`) reads
`oeb_book.opts` to construct a Calibre `Stylizer`. In Calibre's output-plugin
pipeline (`OutputFormatPlugin.convert(oeb_book, output_path, input_plugin, opts, log)`)
the `opts` are passed as a **separate argument** and the `OEBBook` has no `.opts`
attribute (observed on Calibre 9.8). So Stylizer construction raised, the resolver
returned `None`, and **every block got `block_style = None`** — meaning block-level
CSS (`text-align`, `text-indent`, `margin-left/right`, and, as of #15,
`font-family`) was silently dropped and paragraphs fell back to generator defaults.

This was fixed for the font path as part of #15 (commit adding `_ensure_oeb_opts`,
which attaches the conversion `opts` to `oeb_book` so both the #9 resolver and #15
font extraction can build a Stylizer). After the fix, the real conversion emits the
CSS-derived styles and embeds fonts.

## Why a follow-on issue

The `_ensure_oeb_opts` fix repairs the path, but it raises a question the #15 work
did not fully answer: **was the #9 block-styling (CSS typography — alignment,
indent, margins) ever actually verified against a real Calibre conversion, or did
those checks pass on the Calibre-free test shim / native path where the Stylizer is
never constructed?**

The unit/integration tests use `EpubAsOeb` (a Calibre-free shim) or inject a fake
style resolver, so they never exercise real `Stylizer` construction — which is
exactly why this silent failure went unnoticed. See the note in memory:
"Calibre Stylizer signature: Stylizer(tree, path, oeb, opts, profile); only
exercised by a real conversion, not the test shim."

## What to check

- [ ] Confirm block CSS (alignment / indent / margins) now actually renders from a
      **real** `ebook-convert` run on a book that uses those CSS properties — not
      just the shim tests. Compare a real KFX's `$157` styles before/after
      `_ensure_oeb_opts`.
- [ ] Decide whether the earlier #9 device-verification (if any) needs re-running,
      since it may have validated generator defaults rather than source CSS.
[...]

PR #157, opened 2026-09-11
fix: ship image-only pages the TOC never listed; drop only a cover-only page (#152)

### LINKED ISSUE #152: Image-only spine pages the TOC never lists are dropped — a comic with a chapter-level TOC ships 3 of 7 pages
Found converting page-scan books (a comic-style children's series) and checking the output against Kindle Previewer 3.106 on small oracle EPUBs built from the same pages.

## What happens

`_assemble_chapters_by_coordinate` skips any spine item past the last TOC coordinate — and any head before the first — whose text is empty once IMG tokens are removed (the "image-only orphan" rule, 5.3.15). The rule was written for one page: the EPUB's own `cover.xhtml`, whose picture the cover chapter already shows (#32); recovering it produced a blank trailing chapter.

Page-image books are that shape everywhere. A fixed-layout comic, a picture book, any EPUB with one image per spine file and a chapter-level nav has every page after a chapter's first page unlisted.

Measured on a 7-page pre-paginated test EPUB (cover, title, five pages; the nav lists "Cover" and one "Chapter" → page 3):

```
  Skipping image-only orphan p50.xhtml
  Skipping image-only orphan p51.xhtml
  Skipping image-only orphan p52.xhtml
  Skipping image-only orphan back.xhtml
  Chapters: 2          <- 3 of 7 pages reach the reader
```

Kindle Previewer on the same EPUB: 7 sections, 7 images.

A related shape, same cause. A nav-listed "Cover" page holding only the cover image is *not* an orphan, so it becomes a chapter — whose one token the generator cannot resolve, because the cover is excluded from body images (#32). The result is a blank page with a nav entry pointing at it. Publisher EPUBs list "Cover" routinely.

## Suggested

Keep the exception and drop the rule: a page is content unless it shows nothing but the cover image. `extract_cover_image` already knows the cover href; thread it to chapter assembly and compare tokens by basename, the key the generator matches resources with. Recovered pages stay `_omit_from_toc` (#143), and the generator already omits the heading for an image-only body (#33). Apply the same test to a TOC-listed slice, so a cover-only "Cover" chapter is not emitted either.

## Priority

High for page-image books — silent loss of pages. No effect on prose: a text-bearing orphan already recovered.

kmnkv1:
Reported from a page-image book: a fixed-layout comic with a chapter-level
nav lost every page after a chapter's first — "Skipping image-only orphan"
four times on a seven-page test EPUB, three pages reaching the reader.

The rule (5.3.15) skipped any spine item past the last TOC coordinate, or
before the first, whose text was empty once IMG tokens were removed. It was
written for one page — the EPUB's own cover.xhtml, whose picture the cover
chapter already shows (#32) — and a picture book is that shape on every
page.

Keep the exception, drop the rule. `_page_shows_something` treats a page as
content unless it shows nothing but the cover image: `extract_cover_image`
already knows the href, and it is threaded through
`extract_chapters_from_oeb` to `_assemble_chapters_by_coordinate`, where
tokens are compared by basename — the key the generator matches resources
with. The same test sits in `_mk`, so a nav-listed "Cover" page holding only
the cover image is not emitted as a chapter either: as one, its only token
failed to resolve (the cover is excluded from body images) and the reader got
a blank page with a nav entry pointing at it.

Recovered pages keep the existing treatment: unlisted in the nav (#143) and
no invented heading — the generator already omits one for an image-only body
(#33).

    seven-page pre-paginated test EPUB, nav = Cover + one chapter
      before   Chapters: 2     pages shipped 3 of 7
      after    Chapters: 6     pages shipped 7 of 7
    Kindle Previewer 3.106, same EPUB: 7 sections, 7 images

    226-image page-scan EPUB (nav: Cover, front matter, chapters, back matter)
      before   backcover.html skipped as an orphan
      after    225 of 225 referenced images ship

Re-measured end-to-end through `convert_oeb_to_kfx` — image optimisation and
the #102 prune included, output counted out of the KFX rather than out of the
chapter list — over a 226-EPUB library of illustrated children's books
(224 convert; two fail identically on both builds, one malformed XHTML and one
missing zip entry):

                            main        this PR
    image entries ($271)    17,762  ->  17,970    (+208)
    image resources ($164)  14,597  ->  14,800    (+203)
    sections ($260)          5,876  ->   6,013    (+137)
    books gaining images        58   losing 0

The 57 books that lose a chapter earlier in the pipeline lose no image
end-to-end: 4,790 image entries before and after, 1,622 sections down to
1,565. Each of those 57 is a nav-listed cover page whose only token the
generator could never resolve — one blank section per book, gone.

Two existing tests asserted the old rule with an image named cover.jpg and no
cover href; they now pass the href, which is what the production call does.
Prose is unaffected: a text-bearing orphan already recovered, and a page
with any other picture on it now does too.

Device-checked on a Kindle Colorsoft: back cover present, no blank page
after the cover.

Closes #152.


jloutsch:
Thanks for this. I've been trying to reproduce #152 against something I can keep, and I can't — which is itself useful information, so I'd rather ask than guess.

The condition needs a book whose navigation lists **some** spine documents but not all. I tried four: a CBZ through `ebook-convert` gives every page an NCX entry; one fixed-layout sample lists all ten pages; another has thirteen spine documents and an NCX with zero entries, which takes the no-TOC fallback and keeps everything. None of them orphan a page. The corpus can't help either — 0 of its 90 books are fixed-layout.

Could you run this against the book behind the issue and paste the output?

It reports shape only — no title, author, filenames or text, with text measured rather than read — so the output is safe to paste here as it stands. Standard library only, no virtualenv needed. The two numbers I'm after are `spine documents` and `spine docs NOT listed in nav`; if the second is non-zero I can build a fixture that matches and this stops depending on anyone's private files.

kmnkv1:
Ran it. The condition is common in illustrated children's books — publisher EPUBs where the nav lists chapters and the art pages between them are just pages. A Gutenberg corpus can't show it, which I think is the whole reason this looked unreproducible.

The book behind the issue:

```
EPUB shape
==========
  spine documents                      19
  nav entries                          16
  spine docs listed in nav             16
  spine docs NOT listed in nav         3
  fixed layout (pre-paginated)         False
  pages that are mostly text           2
  pages that are image only            16
  spine docs that ARE an <svg>         0
  pages with <svg> inside them         1
  <image> elements inside <svg>        1
  <img> elements                       224
  <img> with width/height attribute    3
  <img> with width/height in style     0
  spine media types                    application/xhtml+xml x19
  image files by extension             jpg x226
```

A stronger one from the same library — 82 of 104 spine documents unlisted, every page image-only:

```
  spine documents                      104
  nav entries                          22
  spine docs listed in nav             22
  spine docs NOT listed in nav         82
  pages that are image only            104
  <img> elements                       103
```

On that book, current `main`:

```
  Skipping image-only head before first TOC anchor
  Skipping image-only orphan part0097.html
  ... part0098 ... part0099 ... part0100 ... part0101 ... part0102
  22 chapters, 96 images
```

with this PR: 29 chapters, 104 images, no drops.

I then ran the whole library rather than one book — 226 EPUBs, mostly illustrated children's series — through `extract_chapters_from_oeb` on `origin/main` and on this branch, via `tests/fixtures/oeb_shim.EpubAsOeb`:

```
  books with unlisted spine docs   202 of 226   (5,661 documents)
  fixed layout                      13
  pages dropped by main            200  in 64 books   ("image-only head"/"image-only orphan")
  pages dropped by this branch       0  — only 23 cover-only pages, in 23 books

  image tokens    18,365 -> 18,642   (+277)
  chapters         5,670 ->  5,836   (+166)
  books gaining      94       losing 58
```

The 58 losers are worth stating explicitly: each loses exactly one token, and I checked all 58 — in every case it is the cover image, on a nav-listed cover page that used to become a chapter the generator could not resolve (the cover is excluded from body images by #32, so that chapter rendered blank). The cover itself still ships through the cover chapter.

Happy to hand over the generator for the synthetic fixed-layout book in the issue — it builds an EPUB of any of these shapes from any images, so a fixture needn't depend on anyone's files. Say the word and I'll open it as a separate PR under `tests/fixtures/`.

jloutsch:
That's the shape I couldn't find, and the reason is now obvious: I was looking at comics and fixed-layout samples, and the common case is an illustrated children's book that isn't fixed-layout at all. 13 of your 226 are FXL; 202 have unlisted spine documents. I was testing the wrong axis.

I reproduced the part of your numbers that looked like a cost — the 58 books losing a token. Building a nav-listed cover page whose only content is the cover image:

| | chapters | image tokens |
|---|---|---|
| `main` | 2 | 1 |
| this branch (cover known) | 1 | 0 |

Confirmed, and confirmed as a correction rather than a loss: on `main` that page becomes a chapter carrying a token the generator cannot resolve, because #32 excludes the cover from body images. It renders blank. Dropping the page is right, and the cover still ships through the cover chapter. So the -1 in those 58 books is a blank page disappearing.

One caveat on the figures, not a problem with them: `extract_chapters_from_oeb` sits a layer below `convert_oeb_to_kfx`, so the run skips image optimisation and the #102 resource prune. Fine for chapter and token counts — just means they aren't end-to-end numbers, and if you re-run anything, the full path is the one I'd quote in the PR body.

**Yes to the generator, please** — open it as a separate PR under `tests/fixtures/`. It is worth more than this PR is. Every one of these defects has been unreproducible for me, and a builder that constructs any of these shapes from arbitrary images ends that permanently, in a form that can be committed. I have a hand-built fixed-layout fixture on a branch here; if yours covers the same ground I will drop mine rather than keep both.

On ordering: this lands first. #159 alone is destructive on a shape from your own library, and #158 alone is a no-op — I've written up both on their threads. This PR is what makes the other two safe.

kmnkv1:
Generator opened as #166, off `main` and independent of the fix PRs — it builds all five shapes plus the rooted-SVG one from #165, from any images. If it covers the same ground as your fixed-layout fixture, drop yours; if there's a shape it can't build, name it and I'll add it.

One thing it turned up on the way, worth knowing independently of that PR: `MINIMAL_JPEG` cannot carry dimensions. Its DQT declares 67 bytes and carries 113, so a strict marker walk — `native_generator._detect_image_dimensions` — stops before the SOF0 and returns `(None, None)`. So no existing fixture can exercise a size-dependent path. #166 adds `jpeg_of(width, height)` and leaves the constant alone.

Re-measured this PR end-to-end as you asked, through `convert_oeb_to_kfx` with image optimisation and the #102 prune, counting out of the KFX rather than out of the chapter list. 226 EPUBs, 224 convert (two fail identically on both builds — one malformed XHTML, one missing zip entry):

```
                        main        this PR
  image entries         17,762  ->  17,970    (+208)
  image resources ($164) 14,597 ->  14,800    (+203)
  sections ($260)         5,876 ->   6,013    (+137)
  books gaining              58     losing 0
```

The end-to-end figure settles the 58 cleanly: **no book loses an image.** Those 57 books (one of the 58 is among the two that fail to convert) hold 4,790 image entries before and after, and drop from 1,622 sections to 1,565 — one blank section each, exactly as your fixture showed. The token disappears a layer above because the generator was already dropping the entry it could not resolve.

PR body updated with those numbers.

PR #158, opened 2026-09-11
fix: honour the size the markup gives an image — width to $56, height to $57 (#153)

### LINKED ISSUE #153: The size the markup gives an image is ignored — `height="98%"` and `width: 100%` render at the pixel-width rule
Found the same way as the image-only-orphan report: page-scan books, Kindle Previewer 3.106 as the oracle, full kfxlib to decode both sides.

## What happens

#145 sizes every image from its pixel width against a 496px column. That is Amazon's rule for an *unsized* image. When the markup gives a size, Amazon honours it and kfxgen does not — the resolver reads `text-align`, `text-indent` and margins, never `width`/`height`, and the IMG token has no field to carry one.

Test EPUB built from a page-scan book (390×625 pages), one sizing idiom per spine file:

```
markup                                        Amazon $157                    kfxgen $157
<img height="98%">                            $57: 98%                       $56: 100%   (1504px ≥ column)
<img style="height:100%">                     $57: 100%                      $56: 77.2%
img.w100 { width: 100% }                      $56: 100%                      $56: 78.6%
<img width="200">                             $56: 39.063%   (= 200/512)     $56: 78.6%
<img style="width:50%">                       $56: 50%                       $56: 78.6%
<img style="width:10em">                      $56: 31.25%    (= 160/512)     $56: 78.6%
<img width="390" height="625">                $56: 76.172%   (width wins)    $56: 78.6%
<img style="max-width:100%;max-height:100%">  $56: 71.094%   (max-* ignored) $56: 73.4%
```

So the rules Amazon applies: an explicit width becomes `$56` — a length divided by the column, a percentage as is, capped at 100; an explicit height becomes `$57`, and only when no width is given; `max-width`/`max-height` are ignored. `$546: $377` accompanies an explicit size.

Page-image books depend on this. A scanned children's book says "fill the page" with `` on every front-matter page; under the pixel rule a 390px scan is three quarters of the column, and a 1504px one is full width but not full height.

**Also measured, not proposing a change:** the column Amazon divides by is 512px on all five test books, less any horizontal *padding* on the image's ancestors — `padding: 0 1em` gives 480, `padding: 0 8px` gives 502.4 — and margins do not count. #145's 496 is consistent with 8px of padding somewhere in that corpus. Worth a look at those seven books; the difference is three percent of width.

## Suggested

Read the computed `width`/`height` of the `` — Calibre's Stylizer folds the HTML attributes into the element's CSS and strips them, so the CSS is the one place both spellings meet — and carry it on the IMG token as an optional third field (`w=98%`, `w=200px`, `h=50%`). In the generator an explicit size wins over `_classify`: width → `$56`, height → `$57` (new `s_img_h<n>` styles alongside `s_img_w<n>`), lengths divided by `IMAGE_COLUMN_PX`, em at 16px. Two-field tokens keep matching.

## Priority

Medium; high for page-image books.

kmnkv1:
Reported from a page-scan book whose three full-page plates are
`` and rendered at three quarters of the column, because
#145's pixel-width rule was the only rule there was. It is Amazon's rule for
an unsized image; when the markup gives a size, Amazon honours it.

Recovered from Kindle Previewer 3.106 on a test EPUB with one sizing idiom
per spine file, both sides decoded with full kfxlib:

    markup                                Amazon            kfxgen (before)
                        $57: 98%          $56: 100%   (1504px >= column)
                 $57: 100%         $56: 77.2%
    img { width: 100% }                   $56: 100%         $56: 78.6%
                         $56: 39.063%      $56: 78.6%   (= 200/512)
                   $56: 50%          $56: 78.6%
                  $56: 31.25%       $56: 78.6%   (= 160/512)
            $56: 76.172%      $56: 78.6%   (width wins)
              $56: 71.094%      $56: 73.4%   (max-* ignored)

Three rules, all applied now: an explicit width becomes $56 — a length
divided by the column, a percentage as is, capped at 100; an explicit height
becomes $57, and only when no width is given; max-width / max-height are
ignored.

The converter reads the computed width/height of the  — Calibre's
Stylizer folds the HTML attributes into the element's CSS and strips them,
so the CSS is the one place both spellings meet; without a Stylizer the
attributes are read directly — and carries it on the IMG token as an
optional third field (`w=98%`, `w=200px`, `h=50%`). A two-field token still
matches; the group is None. In the generator `_image_size_key` puts an
explicit size ahead of `_classify`: widths join the `s_img_w<n>` styles,
heights get `s_img_h<n>` — the "sized" shape with $57 in place of $56 and
the max-width guard kept, so a tall page image cannot overflow the column.
Lengths are divided by IMAGE_COLUMN_PX; em at 16px, pt at 4/3px. A height
given as a length has no reference output and falls back to the pixel rule.

After: identical to Previewer's $56/$57 on every image of the five oracle
books.

How much this actually moves, measured rather than asserted — a 226-EPUB
library of illustrated children's books, every image compared against what
the pixel rule would have emitted:

    image tokens                        18,642
      carrying a size from the markup    6,606   in 131 books
        same $56 as the pixel rule       6,098   <- the attribute restates the natural width
        different $56                      327
        height -> $57 instead of $56       163

So 490 images in 102 books change, not every sized one. They are the ones
where it shows: full-page plates at 76% of the column instead of 98% of the
page height, and images an author deliberately sized down.

Measured and left alone: the column Amazon divides by is 512px on all five
books, less any horizontal *padding* on the image's ancestors
(`padding: 0 1em` gives 480; margins do not count). #145's 496 is
consistent with 8px of padding in that corpus; IMAGE_COLUMN_PX stays until
those books are looked at. Three percent of width either way.

Device-checked on a Kindle Colorsoft: the height="98%" front matter of a
226-image book fills the page height.

Closes #153.


jloutsch:
One thing I can't source: a book that exercises this.

None of the fixed-layout EPUBs I tested carry a sizing hint at all — zero `` elements with a width/height attribute, and zero with width/height in a `style` attribute, across four samples. A CBZ through `ebook-convert` doesn't emit one either; its only stylesheet is an `@page` block zeroing margins, and the `.calibre1`/`.calibre2` classes it writes have no rules. So `height="98%"` is coming from a different producer than calibre's comic path, and I'd like to know which.

Could you paste the shape of the book you measured against?

Shape only — no title, author, filenames or text. The lines that matter here are ` with width/height attribute` and ` with width/height in style`.

Separately, and only so it's written down somewhere: on a synthetic fixed-layout book this PR is a no-op on its own, because the pages carrying the sized images are the ones #157 rescues. With #157 it does what it says. That's an ordering note, not an objection.

kmnkv1:
The producer is a publisher EPUB, not calibre — the book carries `page-template.xpgt` and Adept metadata, i.e. an Adobe/InDesign production chain. Its three full-page plates (two front matter, one back cover) are `Image`; the 224 in-chapter illustrations carry no size. Shape:

```
  spine documents                      19
  pages that are image only            16
  <img> elements                       224
  <img> with width/height attribute    3
  <img> with width/height in style     0
```

Across the whole library I measured against — 226 EPUBs, mostly illustrated children's series:

```
  books with <img> width/height attribute   94 of 226   (6,505 images)
  books with width/height in a style attr   34 of 226   (1,127 images)
```

And a correction I should make before you spend review time on this. I checked what the change actually does to the emitted `$157`, per image, against the natural pixel width:

```
  sized tokens reaching the generator   6,606
    same $56 as the pixel rule          6,098   <- attribute restates the natural width
    different $56                         327
    height -> $57 instead of $56          163
```

So the real behavioural change is 490 images in 102 books, not "every sized image". The PR body oversells it and I'll rewrite that paragraph to these numbers. The 490 are where it shows, though — full-page plates that are 76% of the column instead of 98% of the page height, and images an author deliberately sized down.

Your ordering note is right, and the same interaction bites #159 harder — I've written it up there.

jloutsch:
Thanks for correcting that unprompted — 490 images in 102 books is the number to put in the body, and it being two orders below the headline is the kind of thing that only ever surfaces if the author goes looking. The 6,098 that restate the natural width are the interesting half, incidentally: they say the attribute usually agrees with the pixel rule, which is a point in favour of the #145 arithmetic.

One thing to fix before this merges, which I should have raised when I first read the diff rather than in a request for shape data.

The token gains an optional third `\x01`-delimited field, but `_make_img_token` strips only `\x02` and spaces from alt text — `\x00` and `\x01` pass through. So alt text can open the size field:

```python
_make_img_token('p.jpg', 'photo\x01w=100%')
# -> '\x00IMG\x01p.jpg\x01photo\x01w=100%\x00'
# -> groups ('p.jpg', 'photo', 'w=100%')
# -> _parse_size_hint(...) == ('w', 100.0)
```

Executed, not reasoned. The effect is bounded — an attacker-chosen image width, plus the alt text silently truncating at the `\x01` — so this is not dramatic. But `SECURITY.md` states every field of an EPUB is attacker-controlled, and this adds a new crossing of that boundary where there wasn't one. Stripping `\x00` and `\x01` alongside `\x02` in `_make_img_token` closes it, and it wants a test with a `\x01` in the alt text so it stays closed.

Worth noting the same input reaches a real reader path: a book whose alt text happens to contain a control character would lose part of its alt text and pick up a bogus size, with no error anywhere.

Ordering: #157 first, then this. You're right that it's inert until those pages exist.

kmnkv1:
Fixed in 039ef36, and thank you — I had not thought about the token's own delimiters when I added a field to it.

Executed before and after. The same value also crosses the boundary on the href side, which I don't think the review caught:

```python
_make_img_token('p.jpg\x01w=100%', 'alt')
# before -> groups ('p.jpg', 'w=100%', 'alt')
# after  -> groups ('p.jpgw=100%', 'alt', None)
```

and a NUL in either value is the worse of the two, predating the size field: it ends the token early and the remainder escapes into the text stream as raw control bytes — the #133 shape.

```python
_make_img_token('p.jpg', 'photo\x00tail')
# before -> findall [('p.jpg', 'photo')]   leftover 'tail\x00'
# after  -> findall [('p.jpg', 'phototail')]  leftover ''
```

`_strip_token_controls` removes all three delimiters from href and alt before either is written between them. Two `str.replace` passes per image; alt loses control characters it should never have carried, and nothing else moves. Five tests: separator in alt, separator in href, NUL leaving no remainder, the space escape still stripped from input, and an ordinary sized token unchanged.

Also updated the body to the measured 490, as promised. #159 is rebased onto this so it carries the fix too.

jloutsch:
Housekeeping note, not a review comment — nothing to change here.

`_jpeg` in `tests/unit/test_image_markup_size.py:80` is the fifth way this repo builds a test JPEG. There were already two hand-rolled builders (`test_image_optimize.py`, `test_reproducible_output.py`) and two constants in `tests/_helpers.py`; #160 adds a sixth, byte-identical to yours.

Yours is well-formed, which is worth saying since most of them are not: its COM length of `0x66` correctly covers the two length bytes plus its hundred of payload, and `_detect_image_dimensions` reads it back at 139 bytes. `MINIMAL_JPEG` in `tests/_helpers.py` is the one that is wrong — its padding sits *inside* the DQT without widening that segment's declared length, so a parser following lengths never reaches the SOF0 it does have and reports no size at all. That is the whole of #163: a cover golden that could not see a cover change.

A branch of mine consolidates the two older builders onto a single `make_jpeg(width, height)` in `tests/_helpers.py`, and documents `MINIMAL_JPEG`'s defect at its definition so nobody else reasons from it. I have deliberately not touched your file — rebasing a PR under review to chase a helper is a waste of your time, and the divergence is not what this PR is about.

Once this and #159 are in, I will migrate both copies in one pass. Mentioning it only so that if you touch that helper again you know where it is going.

kmnkv1:
Useful, and one thing to head off before you write it: **#166 already adds that helper** — `jpeg_of(width, height)` in `tests/_helpers.py`, same purpose, same signature. It landed there because the shape fixtures need images that report their own size, and I hit `MINIMAL_JPEG`'s defect on the way without recognising it as #163.

Rather than both of us adding one: tell me which name you want and I'll conform in #166 — `make_jpeg` is fine by me, and yours is the name that already has two call sites waiting to migrate. Or drop the helper from #166 entirely and I'll rebase onto your branch once it's up.

Your numbers are right and mine were not: the DQT carries **116** against a declared 67, not the 113 I wrote, and the SOF0 sits at 138. Measured this time rather than counted. Corrected in 3c7d56e, which also records the defect at the constant's own definition — happy to drop that hunk if your branch does the same thing better.

The duplicate `_jpeg` in this PR's test file I'll leave where it is, as you suggest, and migrate in your pass.

kmnkv1:
Rebased onto the merge of #157. All gates green here: 808 passed, tier3_strict 13 byte-identical, ruff clean. No content change — the rebase was clean.

kmnkv1:
Rebased onto the merge of #166, and this file's `_jpeg` is gone with it — `tests/_helpers.jpeg_of` is the same bytes and is now on main, so two of the six builders you counted are collapsed already. No behaviour change: 18 tests, same assertions.

Gates: 820 passed, tier3_strict 13 byte-identical, ruff clean.

jloutsch:
Rebase verified independently: 823 passed here, tier3_strict 13 byte-identical, ruff clean.

You are right that dropping `_jpeg` collapses two of the six — and it does better than that. Checked across all four of your rebased branches: the only definitions left anywhere are the two pre-existing ones on `main`, which #174 removes. One builder in the repo after that.

It also removed the reason I was holding back a structural guard against this recurring — I did not want to turn your PRs red for something I introduced. That no longer applies, so the guard goes in after #174.

Nothing outstanding on this one. It lands next.

Issue #117, opened 2026-08-22
Illustrations inside a discarded Contents section are dropped with it


jloutsch:
Split out of #113. That fix recovered images dropped by the block walker; this is a separate, narrower loss with a different cause, found by the same corpus assertion.

## Symptom

Two corpus books emit fewer `$164` resources than their source displays inline, after #113 is fixed:

| Book | inline `<img>` refs | resources | lost |
|---|---|---|---|
| pg37106 | 206 | 204 | `b003.png`, `b004.png` |
| pg1400 | 32 | 31 | `0012m.jpg` |

## Where it happens

Not in extraction, and not in the block walker. Traced layer by layer:

- `extract_images_from_oeb` finds **all** of them (32/32, 206/206)
- `extract_blocks_from_html` produces an IMG token for **all** of them
- `extract_chapters_from_oeb` is where they disappear

## The shape

All the lost images live on the Gutenberg front-matter page, which is *not* discarded as a whole — it carries 17,878 characters of text and six other images on it survive. Comparing them on that one page:

```
kept   b001.png   after <h1>LITTLE WOMEN.</h1>
kept   b002.png   after the printer's colophon
LOST   b003.png   after <h2 id="pgepubid00001">Contents</h2>
LOST   b004.png   after the Contents <table>
kept   b005.png   after <h2 id="pgepubid00002">Illustrations</h2>
kept   b006.png   after the Illustrations <table>
kept   b007.png   after <h2>I. Playing Pilgrims.</h2>
```

The two that vanish are exactly the two between the `Contents` heading and the next heading. That is the region the converter drops as a redundant table of contents (#76 records the same behaviour from the other direction: a fixture page titled "Contents" never runs because those blocks are discarded upstream).

So the rule is: discarding the TOC section takes any illustration inside it along. `b003.png` and `b004.png` are decorative plates that happen to sit there, not navigation.

## Two ways to read this

**It is a bug.** A reader loses two plates that the print edition has. Nothing warns; the file is structurally valid and every emitted resource is referenced.

**It is correct.** KFX carries its own navigation, the source TOC is redundant, and dropping the section wholesale is the documented behaviour. The images are collateral, and preserving them means reaching into a section that is otherwise deliberately ignored.

I lean toward the first for images specifically — a plate is content, and the reason to drop a TOC section is that its *text* duplicates navigation, which says nothing about pictures in it. But it is a judgement call about intent, so it is worth deciding rather than assuming.

[...]

PR #159, opened 2026-09-11
fix: read <svg><image> pages as images (#154)

### LINKED ISSUE #154: `<svg><image>` pages come through empty
## What happens

`extract_blocks_from_html` recognises `` and nothing else. A spine file whose picture is

```html
<div><svg viewBox="0 0 1682 2475"><image width="1682" height="2475" xlink:href="../images/cover.jpg"/></svg></div>
```

yields no text and no token, so `extract_chapters_from_oeb` skips the item entirely and "TOC entry 'Cover' dropped: not in spine" follows.

That wrapper is not exotic. calibre's own EPUB output writes `titlepage.xhtml` exactly this way, and publisher EPUBs use it for the cover and for full-page plates — the example above is a Wimpy Kid EPUB's `cover.xhtml`.

Kindle Previewer 3.106 renders them. Mid-book, an SVG page becomes a `$270` block sized `$56: 100%, $57: 100%` holding the image at `$56: 100%`; as the first spine page it becomes the fixed-size cover section. Either way the picture reaches the reader.

## Suggested

Treat `<svg>` as an image everywhere the walker treats `` — leaf block, inline run, container dispatch, discarded listing — reading `xlink:href`/`href` of each `<image>`. Size it at height 100%, which in the reading flow is the same look as Amazon's 100%×100% block; that needs the size field on the IMG token from the markup-size issue.

## Priority

Medium. Every calibre-made EPUB has one such page; most of the time it is the cover, which #32 rescues separately, so the visible damage is on books that wrap plates this way.

kmnkv1:
A spine file whose picture is an <svg> wrapping an <image> — calibre's own
titlepage.xhtml, a publisher's cover.xhtml, full-page plates — yielded no
text and no token: `extract_blocks_from_html` knew  and nothing else.
The item was skipped, and the nav entry pointing at it dropped ("TOC entry
'Cover' dropped: not in spine").

Kindle Previewer 3.106 renders them. Mid-book an SVG page becomes a $270
block sized 100%x100% holding the image at $56: 100%; as the first spine
page it becomes the fixed-size cover section.

<svg> is now an image at every point the walker treats : the inline
run, the container dispatch, the leaf block, and a discarded contents
listing (#117 keeps pictures there). Each <image> it draws — xlink:href or
href — becomes a token sized `h=100%`, which in the reading flow is the same
look as Amazon's block. Builds on the size field from #153; that commit is
included here.

    test EPUB with an SVG cover page and two SVG plate pages
      before   all three spine items empty, "TOC entry 'Cover' dropped"
      after    all three carry their image token; the cover page shows only
               the cover and is handled by the cover chapter (#32, #152)

Closes #154. Builds on #158 — the first commit here is that change; rebases to one commit once it lands.


jloutsch:
A question about scope, because I think there are two shapes here and this PR handles one of them.

Testing against an EPUB 3 fixed-layout sample whose spine items are `image/svg+xml` — the page *is* an SVG file rather than an XHTML page containing one — the converter extracts zero blocks per page, and this branch doesn't change that. The root element is `{svg}svg`, and `extract_blocks_from_html` walks from `<body>`, which such a document doesn't have. Six spine documents, six `<image>` elements, nothing emitted, before and after.

The nested form this PR does fix is real and does occur: calibre's comic input generates a `titlepage.xhtml` wrapping the cover in exactly the markup you handle, with `height="100%"` on the `<svg>` — which incidentally supports the `h=100%` constant you chose.

So: are the pages in the book behind #154 nested `<svg>` inside XHTML, or spine documents that are themselves SVG? The shape report tells the two apart:

Compare `spine docs that ARE an <svg>` against `pages with <svg> inside them`. Shape only — no title, author, filenames or text. If it's the rooted form, this PR needs to grow a case for it; if it's the nested form, the rooted case is a separate issue and I'll file it.

kmnkv1:
Nested, in every case. Across the 226 EPUBs I have:

```
  pages with <svg> inside them        104 of 226 books
  spine docs that ARE an <svg>          0 of 226 books
```

So the rooted form is out of scope here — please file it separately, and I'll take a look if you like.

One thing worth knowing before this merges, though: **this PR alone is a net regression on that library, and #157 is what makes it safe.**

```
  main                18,365 image tokens
  + this PR (alone)   18,012   (-353)
  + #157 as well      18,642   (+277)
```

Two books lose 225 of 226 images each. Their shape:

```
  spine documents                       2
  nav entries                           1
  spine docs listed in nav              1
  spine docs NOT listed in nav          1
  pages with <svg> inside them          1
  <img> elements                      226
```

The spine is `[titlepage.xhtml (the SVG cover), content.xhtml (226 images)]` and the NCX points only at the titlepage. Today the SVG page yields no blocks, so it never enters the spine list and the content document is the only coordinate. Once the SVG is extracted the cover becomes the *first* coordinate, and the 226-image document falls past the last one — so it is dropped as an image-only orphan:

```
  Skipping image-only orphan index.html
  1 chapter, 1 image
```

With #157 in, both books are whole again (226 images). Two spine items and a one-entry NCX is a common shape for a PDF-derived book, so I don't think it's exotic. Merging #157 first, or both together, avoids it entirely.

And thanks for checking the `height="100%"` on calibre's titlepage — that's better provenance for the `h=100%` constant than what I had.

jloutsch:
I rebuilt the shape you described — spine of `[svg titlepage, content document]`, NCX pointing only at the titlepage — and it reproduces exactly as you reported:

| build | image tokens |
|---|---|
| `main` | 226 |
| **this PR alone** | **1** |
| #157 alone | 226 |
| #157 + this PR | 227 |

225 of 226 images lost, and your account of why is right: the SVG page yields no blocks today, so it never becomes a coordinate; extracting it makes the cover the first coordinate and drops the content document past the last one. Thank you for finding that and saying so on your own PR — it would have shipped otherwise, and it is the kind of thing a corpus of prose could never have caught.

So this does not land alone. #157 first, or both together.

The rooted case is filed as #165, with the `IDPF/epub3-samples` book as its reproducer so it doesn't depend on anyone's library. I noted your 0-of-226 measurement there as the reason it's low priority — worth fixing when the SVG path is next opened rather than on its own. If you want it, take it; if not it can sit.

Your `h=100%` constant stands. Calibre writes `height="100%"` on the `<svg>` in the titlepage it generates for every comic, so the constant matches the producer rather than just working.

kmnkv1:
Rebased onto the merge of #157, so the ordering constraint is satisfied and the destructive interaction is gone. Re-measured across the 226-EPUB library against the new `main`:

```
  main (with #157)   18,534 image tokens   5,806 chapters
  + this PR          18,642 (+108)         5,836 (+30)
  books gaining 68     losing 0
```

Where the old figure was −353 with two books losing 225 images each, it is now +108 with nothing lost anywhere. Biggest single gain is 13 images in one book whose plates are all `<svg>`-wrapped.

Gates: 813 passed, tier3_strict 13 byte-identical, ruff clean. Carries the token-delimiter fix from #158 as its base commit, as before.

kmnkv1:
Rebased onto the merge of #166, through #158. Gates: 825 passed, tier3_strict 13 byte-identical, ruff clean.

jloutsch:
Rebase verified: 828 passed here, and the SVG work itself is sound.

**One thing from the original review is still open, and it is the only thing holding this back.** An `<image>` inside `<defs>` is still painted:

```python
<svg><defs><image href="hidden.jpg"/></defs><image href="a.jpg"/></svg>
# hidden.jpg emitted: True
```

`_svg_image_refs` walks `svg.iter()`, which descends into `<defs>`, `<symbol>`, `<mask>`, `<clipPath>` and `<pattern>`. SVG paints none of those — an `<image>` in a `<defs>` is a definition, drawn only where a `<use>` references it. So a picture the publisher hid is drawn, and the resource is embedded rather than pruned by #102.

Worth noticing that you have already solved the same class of problem next door: #167 now returns zero blocks for a rooted `<svg>` carrying a `<style>`, where before it emitted the stylesheet as body text. I verified that on the sample — `cover.svg` went from 1894 characters of CSS to nothing. This is the same shape one level down.

I have the fix and a golden fixture on a branch here — `_walk` that prunes the non-rendered containers, plus an `svg_images` fixture whose decoy is in the manifest so that painting it shows up as a fourth resource rather than being dropped as unresolvable. Happy to push it to you, or to open it against your branch, or you take it — whichever you prefer. It is about ten lines.

That is the last item on this one.

kmnkv1:
Took it — third option, since the code is in hand. If your golden fixture differs from what this does, push it on top and I will adopt it rather than argue about ten lines.

`_svg_image_refs` walks now instead of `iter()`, so a subtree that SVG does not paint is skipped whole:

```
<svg><defs><image href="hidden.jpg"/></defs><image href="a.jpg"/></svg>
  before   hidden.jpg, a.jpg
  after    a.jpg
```

`defs`, `symbol`, `mask`, `clipPath`, `pattern`, `marker` — `<g>` and `<a>` still walked, which is the control test.

You are right that it is the same shape as the leak in #167, and the same answer: a container that is not drawing contributes nothing. I have written the `<use>` consequence into the comment rather than leaving it implicit — an image defined in one of these and referenced elsewhere is now dropped rather than painted in the wrong place, because `<use>` is not resolved. Dropping is the lesser error, and a reference file with a `<use>`-referenced image is what would change that.

Nine tests: one per container, a definition nested two levels down, `<g>`/`<a>` as the control, and a page whose only image is a definition drawing nothing.

Gates: 836 passed, tier3_strict 13 byte-identical, ruff clean. Rebased onto the merge of #158, and #167 re-stacked on top of this — it picks the rule up for rooted SVG documents too.

jloutsch:
You had already fixed this by the time I came to push my version, so nothing from me is needed — and your set is better than mine was. I had `defs, symbol, mask, clipPath, pattern`; yours adds `marker` and lowercases before matching, which handles `clipPath` written either way.

Verified rather than taken on trust:

```
defs     non-rendered emitted: none   a.jpg drawn: True
mask     non-rendered emitted: none   a.jpg drawn: True
symbol   non-rendered emitted: none   a.jpg drawn: True
```

And the tests discriminate — reverting only the container skip, leaving everything else in place:

```
8 failed, 6 passed
FAILED TestNonRenderedSvgContainers::test_a_page_whose_only_image_is_a_definition_draws_nothing
```

A parametrized sweep over all six containers plus the nested-in-`<g>` case is more than I would have written.

Gates here: 839 passed, tier3_strict 13 byte-identical, ruff clean.

**Nothing outstanding on this PR.** It was the last item in the fixed-layout chain that needed anything from either of us. I am dropping my `svg-golden-on-159` branch — it is a subset of what you have.

PR #160, opened 2026-09-11
fix: lay the cover section out as a fixed-size block, as Amazon does (#155)

### LINKED ISSUE #155: Amazon lays the cover out as a fixed-size block section; kfxgen emits an inline section holding a 100%-wide image
## What happens

`build_fragment_260`'s docstring already knows the shape — "Block display ($270) with fixed dimensions causes text overlap and is only appropriate for cover image sections" — but every section, the synthetic cover chapter (#32) included, is emitted as `$159: $269` with the cover as an ordinary `s_img_w<n>` image at 100% width.

Every Kindle Previewer 3.106 build whose first spine page is the cover image — five test books, `` and `<svg>` covers alike, and calibre's own comic EPUB — has this instead:

```
$260 c0: {$155: 918, $176: l4, $66: 1495, $67: 2200, $156: $326, $140: $320, $159: $270}
$259 l4: {$176: l4, $146: [{$155: 873, $157: s8, $159: $271, $175: e6}]}     s8 carries no width
```

`$66`/`$67` are the cover's pixel size; the image entry inside carries no width style. The device scales the block to the screen, so the cover fills the page rather than sitting inside the text margins with the reader's margin setting around it.

## Suggested

When the section is the `_is_cover` chapter and the cover's dimensions are known (`_detect_image_dimensions`), emit that entry; otherwise unchanged. Keep the image entry's style — harmless inside a block sized to the image.

Pure rendering change, invisible to tiers 1–3, so it wants a device look. Checked on a Kindle Colorsoft on a 226-image page-scan book: the cover fills the page, no blank page after it, navigation unaffected.

## Priority

Low–medium. Cosmetic, but it is the first page of every book.

kmnkv1:
`build_fragment_260`'s docstring already knew the shape — block display with
fixed dimensions "is only appropriate for cover image sections" — but every
section, the synthetic cover chapter (#32) included, was $159: $269 with the
cover as an ordinary 100%-wide image inside the text margins.

Every Kindle Previewer 3.106 build whose first spine page is the cover image
(five test books,  and <svg> covers, and calibre's own comic EPUB):

    $260 c0: {$155, $176: l4, $66: 1495, $67: 2200, $156: $326, $140: $320, $159: $270}

$66/$67 are the cover's pixel size; the device scales the block to the
screen. The `_is_cover` chapter's section now carries that entry when the
cover's dimensions are known, and is unchanged otherwise — a cover whose
SOF the dimension scan cannot read keeps the inline section. The image
entry keeps its style; inside a block sized to the image it is harmless.

Pure rendering change: tiers 1–3 cannot see it, and the goldens stay
byte-stable (their cover fixture has no readable dimensions), so it was
looked at. Device-checked on a Kindle Colorsoft on a 226-image page-scan
book: the cover fills the page, no blank page after it, navigation
unaffected.

Closes #155.


jloutsch:
Housekeeping note, not a review comment — nothing to change here.

`_jpeg` in `tests/unit/test_cover_section.py:17` is byte-identical to the one #158 adds in `tests/unit/test_image_markup_size.py`, and both are well-formed. Between them, the two existing hand-rolled builders and the constants in `tests/_helpers.py`, this repo will have six ways to make a test JPEG once these land.

That matters more here than anywhere else in the queue, because this PR is the one the divergence already bit. `with_cover` pinned the unknown-size branch — the branch a real book almost never takes — because its cover was `MINIMAL_JPEG`, whose padding sits inside the DQT without widening that segment's declared length, so the dimension scan never reaches the SOF0 it has. Your change altered 90 of 90 corpus books while `test_golden_byte_identical[with_cover]` stayed green. That is #163, now fixed on a branch here, and with the fixture repaired this PR correctly fails that gate.

So your instinct to write a JPEG that reports its own size was right, and the repo's existing fixture was the thing at fault.

A branch of mine adds `make_jpeg(width, height)` to `tests/_helpers.py` and moves the two older builders onto it. I have not touched your file. Once the gate fix and this PR are both in, I will migrate the remaining copies in one pass.

kmnkv1:
Understood, and the gate finding is the useful half of that note.

I did see that fixture was blind — the PR body says the goldens stay byte-stable "(their cover fixture has no readable dimensions)" — and I treated it as a reason the change was safe rather than as a defect in the gate. That was the wrong conclusion to draw from it: a fixture that cannot observe the branch under test is a broken gate, not a quiet one, and 90 of 90 corpus books changing behind a green check is exactly the failure mode. I should have filed it. Thanks for doing so as #163.

So: this PR *should* fail `test_golden_byte_identical[with_cover]` once #163 lands. When it does, say the word and I'll rebase onto it and regenerate `expected/with_cover.kfx` in this branch — or if you'd rather keep the regeneration in the gate-fix commit so the two are never separated, that works too and I'll rebase onto whatever the golden ends up being.

On the JPEG builders: #166 already adds `jpeg_of(width, height)` to `tests/_helpers.py`, which is the same helper you have on your branch as `make_jpeg`. I've said on #158 that I'll take whichever name you prefer rather than land a second one.

Worth noting the device check stands on one model only — a Colorsoft, where the cover fills the page and nothing else moved. Your tier-4 table pins Paperwhite, Oasis and Voyage, and I have none of them, so this wants your hardware before it is a tier-4 pass rather than a report.

kmnkv1:
Rebased onto #169 and regenerated. The gate behaved exactly as you said it would:

```
before #169   test_golden_byte_identical[with_cover]   passed   <- blind
after  #169   test_golden_byte_identical[with_cover]   FAILED   <- correct
after regen   13 passed
```

The regenerated golden differs in the one thing this PR is for and nothing else:

```
  c0 section entry   before   $155 $176 $159
                     after    $155 $176 $66 $67 $156 $140 $159
```

The other twelve are untouched. Your `test_with_cover_fixture_has_a_measurable_cover` guard passes too, so the fixture cannot slide back to pinning the unknown-size branch behind this.

Still device-checked on a Colorsoft only — your tier-4 table wants Paperwhite, Oasis and Voyage, and I have none of them.

kmnkv1:
Rebased onto the merge of #166. This branch's `_jpeg` is dropped in favour of `tests/_helpers.jpeg_of`, which is byte-identical and now on main — the regenerated with_cover golden is unaffected by that.

Gates: 804 passed, tier3_strict 13 byte-identical (with_cover included), ruff clean.

jloutsch:
Rebase verified: 807 passed here, tier3_strict 13 byte-identical including `with_cover`, and the regenerated golden matches what the code emits — the gate passing is evidence of that, not of it being skipped.

Dropping `_jpeg` for `jpeg_of` is right and it closes the count: the only copies left anywhere are the two pre-existing ones on `main`, which #174 removes. After that there is one builder in the repo.

**One thing the regeneration changes, and it is worth naming.** With the golden re-recorded, this PR is green everywhere — and the fact it rewrites every book in the corpus is now visible only in this thread. I re-measured against the rebased branch:

```
books whose bytes change under this PR: 90 of 90
```

That is not an objection. Regenerating is the correct procedure for an intended output change, and the gate ladder puts device sign-off at the release tag rather than at merge, so this can land on the tests alone. But the byte gate has stopped asking the question, so the answer has to come from somewhere else, and for this change that is hardware.

What I would want to see on a device, in the order the defects would show:

1. **The cover fills the page.** That is the claim, and it is the whole reason to take the change.
2. **TOC navigation still lands on the right page.** The code immediately above your edit says explicit `$156: $326` "causes TOC navigation to land one page late (preview shows correct page but tap goes to next page)". You add exactly that property to the cover section. It may well be harmless there — nothing navigates *to* a cover — but the comment is a record of someone having been bitten, and the PR does not say why this case differs. Tapping two or three TOC entries settles it.
3. **A book whose cover has no readable dimensions** still renders, since that path takes the old branch.

If you have a Kindle to hand and want to do that pass, say so and I will take the numbers; otherwise it waits on mine, which is fine — nothing else is blocked behind it.

The `$156: $326` question is the one I would most like answered in words either way, even before hardware: is it there because Amazon emits it on the cover section, or because the block layout needs it?

kmnkv1:
Answering the `$156: $326` question in words first, since it does not need hardware.

**It is there because Amazon emits it, not because the block layout needs it.** Every Kindle Previewer 3.106 build I decoded whose first spine page is the cover image has the property on the cover section, `` covers and `<svg>` covers alike, and one of them is calibre's own comic EPUB:

```
{$174: c0, $141: [{$155: 928, $176: l4, $66: 1495, $67: 2200, $156: $326, $140: $320, $159: $270}]}
{$174: c0, $141: [{$155: 918, $176: l4, $66: 1200, $67: 1600, $156: $326, $140: $320, $159: $270}]}
{$174: c0, $141: [{$155: 934, $176: l4, $66: 1495, $67: 2200, $156: $326, $140: $320, $159: $270}]}
{$174: c0, $141: [{$155: 918, $176: l4, $66: 1495, $67: 2200, $156: $326, $140: $320, $159: $270}]}
```

Four books, same shape. I copied the whole entry rather than assembling one, which is why `$140: $320` is there too. I have not tested whether the block renders correctly *without* `$156`, so I cannot tell you it is load-bearing — only that removing it would be a divergence from the reference rather than a simplification.

On the warning above the edit: the case it records is a *body* section, where something navigates to it. Nothing navigates to a cover — it is never a TOC target, and `_omit_from_toc` is set on the synthetic cover chapter — so the mechanism it describes has nothing to land one page late. That is an argument, though, not a measurement, which is exactly why your check 2 is the right one to run.

The device pass will happen, but not today — the Kindle here is not to hand for a day or two. Please don't hold the queue for it: #159 and #167 are ahead of this and independent of it, and if your own hardware comes free first, take the numbers from that. When I do run it I will report all three checks, including a book whose cover has no readable dimensions taking the old branch.

Rebased onto the merge of #158 and #161 meanwhile: 824 passed, tier3_strict 13 byte-identical including `with_cover`, ruff clean.

jloutsch:
That answers it, and four decoded reference entries is better evidence than I asked for. "Copied the whole entry rather than assembling one" is also the right instinct for a format nobody has a specification for — `$140: $320` riding along is a feature of that approach, not a loose end.

**I have measured the part of your argument that was an argument.** You said nothing navigates to a cover, so the mechanism the warning describes has no target. That is checkable without hardware, and it holds twice over:

```
$389 fragments: 1
nav mentions 'c0': False
```

The navigation fragment does not reference the cover section at all. And the stronger version — whether adding the property shifts anything *else*:

```
                    main                              this PR
section positions   c0..c6 = 10000,10002,...,10012    identical
nav position refs   1006,1014,1022,1030,1038,1046     identical
```

Byte-identical on both. The change is confined to the cover section's own layout entry; it moves nothing in the position map and nothing in the navigation. So the failure mode the comment above your edit records — a TOC entry landing a page late — has neither a target on that section nor a shifted position anywhere else.

That does not make check 2 unnecessary, but it does change what it is for: it is now confirming that rendering behaves, not that navigation still points where it did. The structural risk is measured at zero.

**Your rebase verified independently:** 827 passed here, tier3_strict 13 byte-identical including `with_cover`.

Taking your advice on the queue — #159 and #167 go ahead of this, and #160 waits on whichever Kindle frees up first. Nothing is blocked behind it.

The remaining item on this PR is now one device pass and nothing else.

kmnkv1:
Rebased onto the merge of #159 and the tier-2 work. Gates: 845 passed, tier3_strict 13 byte-identical including `with_cover`, ruff clean.

Thank you for measuring the argument rather than accepting it — "nav mentions 'c0': False" plus identical position maps is a better answer than the one I gave. The device pass is still coming from here; it is a day or two out, so if your hardware frees up first please just take it.

jloutsch:
Device pass run. **The change is not visible on the hardware I have**, and I think that puts this in the same position as #120.

## What was tested

An A/B pair built from the same book — a 37-page public-domain comic, every page a scan, cover 1200x1600 so the new branch actually fires. Retitled per build so the two got distinct ASINs and did not overwrite each other. The only difference between them is the cover section's layout entry:

```
A (main)    $159=$269
B (this PR) $159=$270  $66=1200  $67=1600  $156=$326  $140=$320
```

Kindle Oasis 10th generation, firmware 5.18.2.

| check | result |
|---|---|
| cover appearance | **identical in both** |
| TOC navigation, three entries including the last | identical in both, lands correctly |
| home-screen thumbnail | fails in both — filed separately as #188 |

## What that means

The `$156: $326` worry is settled: nothing landed late, which matches the structural measurement that nav targets and position maps are byte-identical between the two builds. Your argument about nothing navigating to a cover holds.

But the claim this PR exists for — that the cover fills the page instead of sitting inside the text margins — does not reproduce. Both covers look the same. Against that: the change rewrites the bytes of 90 of 90 corpus books.

That is #120's shape exactly. There the rule was worked out and verified to 1,719 of 1,720 images, and then a device pair showed the two builds rendering identically, and it was closed as measured rather than merged — because the change cost an edit in a fragile emission path and bought nothing a reader could see. The reasoning transfers.

## Before I act on that

**One device, one firmware.** The Oasis is on 5.18.2 and terminal; the Paperwhite here runs 5.19.2 and is not to hand. A cover-layout difference between firmware generations is entirely plausible, so this is "no benefit on the stricter device", not "no benefit anywhere". You mentioned a Colorsoft — if you are willing to run the same pair on it, that is the evidence that would change the answer, and it is a different enough device to be worth more than a second Oasis would be.

Happy to send you the two builds, or the recipe: convert one book twice, once per branch, retitling the source each time so the ASINs differ.

**Your decoded reference entries still stand.** Four Previewer builds emitting this shape on the cover section is real, and "Amazon does it" is a decent reason to match. It is just not, on its own, a reader-visible improvement — and this project has an explicit precedent for not merging those.

## Where I would leave it

Open, pending a second device. If the Colorsoft shows the same, I would close it the way #120 was closed: with the measurement recorded, so the next person who has the idea finds out it was tried. Nothing about that is a judgement on the work — the investigation was sound and the reference data is worth keeping either way.

Not blocking anything: it is the last open PR, and #188 is independent of it.

Issue #163, opened 2026-09-19
The with_cover golden pins the unknown-size branch, because MINIMAL_JPEG is malformed


jloutsch:
`MINIMAL_JPEG` in `tests/_helpers.py` is malformed. Its DQT segment declares a length of
67 bytes and carries 116. `NativeKFXGenerator._detect_image_dimensions` follows segment
lengths, as a JPEG parser must, so it walks off the quantization table and never reaches
the SOF0 that sits at offset 138. It returns `(None, None)` for what is nominally a 1x1
image.

The constant's own docstring calls it "a minimal-but-valid JPEG". It is not valid.

## Why it matters here

`make_with_cover` uses it as the cover. The generator branches on whether a cover's
dimensions are known, so that fixture pins the unknown-size branch — the branch a real
book almost never takes. Any change to how a *measurable* cover is laid out is invisible
to it.

Measured against PR #160, which lays the cover section out as a block carrying the
image's pixel dimensions:

| | result |
|---|---|
| `test_golden_byte_identical[with_cover]` | passed |
| Gutenberg corpus books whose bytes changed | 90 of 90 |

The one fixture whose job is to lock cover emission could not see a change that altered
every book in the corpus.

## Fix

Add a well-formed JPEG with readable dimensions and point the cover fixture at it, then
regenerate `expected/with_cover.kfx`. A guard should assert the fixture's cover still
reports its own size, so the fixture cannot quietly return to pinning the unknown-size
branch.

`MINIMAL_JPEG` itself is a wider question — it is used by nine fixtures, and repairing it
would move every body image from the unknown-size bucket into `small` and churn goldens
well past this fix. Tracked separately.

## Verification

With the fixture repaired, applying #160's generator change fails the byte gate at
`test_golden_byte_identical[with_cover]`, which is what it should have done all along.

jloutsch:
## Correction: the DQT numbers above are wrong

The issue body says `MINIMAL_JPEG`'s DQT "declares a length of 67 bytes and carries 116". The 67 is right, the 116 is not. #166 states 113, which is also not. Computed rather than estimated:

```
marker ffdb at offset 20: declared length 67 -> next marker expected at 89
DQT payload starts at 24, real SOF0 found at 138
declared payload  65 bytes   (the length field counts its own two)
actual payload   114 bytes
overrun           49 bytes
```

[...]

PR #166, opened 2026-09-19
test: build the page-image shapes, so these defects stop needing my library

kmnkv1:
Asked for in [#157](https://github.com/jloutsch/kfxgen/pull/157#issuecomment-5738696325). Independent of the five fix PRs — it lands on `main` on its own, and the tests in it assert what the builder emits rather than what the converter does with it, so they hold on any branch.

The reason none of these defects reproduced for you is that the condition is not "a comic" or "fixed layout" — it is a shape. Measured over a 226-EPUB library of illustrated children's books:

```
  spine documents the nav never lists   202 of 226 books   (5,661 documents)
  art wrapped in <svg>                  104 of 226
  a size stated on the <img>             94 of 226         (6,505 images)
  fixed layout                           13 of 226
  spine documents that ARE an <svg>       0 of 226
```

`EpubBuilder` grows those as arguments rather than as new classes:

```python
EpubBuilder()
    .set_fixed_layout(width=390, height=625)
    .set_cover(jpeg_of(1495, 2200), href="cover.jpg")
    .add_image_page("Cover", image=cover, href="cover.jpg")            # shows only the cover
    .add_image_page("Title", image=jpeg_of(1504, 2192),
                    attrs={"height": "98%"})                           # sized by the markup
    .add_chapter("Chapter One", "Body.")
    .add_image_page("Plate", image=jpeg_of(390, 625), listed=False)    # in the spine, in no nav
    .add_image_page("Plate", image=..., wrapper="svg")                 # nested <svg>
    .add_image_page("Page",  image=..., wrapper="svg-root")            # the document IS an <svg> (#165)
    .add_calibre_inline_toc()                                          # calibre's generated contents page
    .build(tmp_path, "page_images")
```

One thing worth flagging separately. `tests/_helpers.jpeg_of(width, height)` is not `MINIMAL_JPEG` with the numbers swapped, because that constant cannot carry dimensions: its DQT declares 67 bytes and carries 114 of payload against the 65 declared, so a parser that follows lengths — which is what `native_generator._detect_image_dimensions` does, as a JPEG parser must — stops at 89 and never reaches the SOF0 at offset 138, 49 bytes further on. That is #163.

```
_detect_image_dimensions(MINIMAL_JPEG)      -> (None, None)
_detect_image_dimensions(jpeg_of(390, 625)) -> (390, 625)
```

Correct for the magic-byte sniff it was written for; it just means no existing fixture can exercise a size-dependent path. `jpeg_of`'s marker chain is honest end to end and both readers in the codebase agree on it. `MINIMAL_JPEG` itself is untouched.

The document list became one ordered structure so pages and chapters interleave in reading order, and numbering counts every spine document — which leaves a book of chapters alone named exactly as before. **tier3_strict passes, all 13 goldens byte-identical.** 12 new tests, full suite green apart from the two Windows-only failures already on `main` (`test_safe_write`).

If it covers the same ground as the fixed-layout fixture on your branch, drop yours — and if there's a shape here it can't build yet, say which and I'll add it.


jloutsch:
Verified the parts that could go wrong quietly: 794 passed, tier3_strict 13 byte-identical, ruff clean. The `EpubBuilder` restructure not moving a single golden byte is the thing I would have expected to cost a day, so thank you for doing it that way round.

Taking it. I have a hand-built fixed-layout fixture on a branch here and it builds a strict subset of what this does — I will drop it rather than keep two.

Two corrections, one of them mine.

**The DQT figures.** You have `MINIMAL_JPEG` declaring 67 and carrying 113; I wrote 116 in #163. Both are wrong, and since this is now the explanation two files and an issue point at, here is the arithmetic:

```
marker ffdb at offset 20: declared length 67 -> next marker expected at 89
DQT payload starts at 24, real SOF0 found at 138
declared payload 65 bytes (the length field counts its own two)
actual payload  114 bytes
overrun          49 bytes
```

So: declares 67, carries 114, and the SOF0 it does have sits 49 bytes past where a length-following parser stops. I will correct #163.

**`test_safe_write`.** You mention "the two Windows-only failures already on `main`". On macOS that file is 22 passed, and CI is ubuntu-only, so nothing here has ever seen them. If they really do fail on Windows I would like them as their own issue rather than a known-quantity aside — `O_NOFOLLOW` is POSIX-only and `native_generator.py` degrades it to a no-op there, so a genuine Windows failure would mean the symlink defense is weaker on a platform nobody tests. Worth noting that CI was cut from twelve jobs to one on the reasoning that the code has "no platform-specific behavior", which that degradation contradicts.

Nothing blocking. One small thing if you are touching it again: `jpeg_of` in `tests/_helpers` now sits alongside a `make_jpeg(width, height)` I added on a branch for the same reason, plus `SIZED_JPEG`/`NARROW_JPEG`. I will collapse those into whichever of ours reads better once this lands — no action for you.

kmnkv1:
Your arithmetic, adopted in 63738ff — both the comment at the constant and the `jpeg_of` docstring now say 114 of payload against the 65 declared, parser stopping at 89, SOF0 at 138, overrun 49. Mine counted the segment span including the length bytes and yours counted to the wrong end; the computed version is the one worth citing.

`test_safe_write` is filed as #173, and the answer is better than either of us assumed: **the defense holds on Windows, both failures are POSIX assumptions in the tests.**

```
test_mode_is_0644                      st_mode is 0o100666 — Windows has no
                                       permission bits and no umask
test_refuses_symlink_at_destination    OSError ELOOP raised          ok
                                       target NOT overwritten        ok
                                       still a symlink               ok
                                       os.readlink() == str(path)    FAILS — \?\ prefix
```

So `O_NOFOLLOW` degrading to `0` does not leave the destination unprotected: the `islink()` check is what rejects, and it fires. What is genuinely weaker without it is the TOCTOU window on the tmp slot, which the docstring already declares out of scope. Suggested fix in the issue is four lines and touches no production code — say the word and I'll send it.

On the helper: take whichever of `jpeg_of` / `make_jpeg` you prefer when you collapse them — no preference here beyond there being one.

PR #167, opened 2026-09-19
fix: read a spine document that is itself an <svg> (#165)

### LINKED ISSUE #165: A spine document that IS an <svg> is never read — extract_blocks_from_html walks from <body>
A spine document whose media type is `image/svg+xml` — the page *is* an SVG file rather than an XHTML page containing one — produces no blocks. Every image it draws is lost.

This is the sibling of #154. That issue is about `<svg>` nested inside an XHTML page, which #159 addresses. This one is about `<svg>` as the spine document's own root element, which #159 does not reach.

## Measurement

Against the EPUB 3 fixed-layout sample `svg-in-spine.epub` from `IDPF/epub3-samples` (freely licensed, so this is reproducible without anyone's private files):

```
  spine documents                      6
  nav entries                          3
  spine docs that ARE an <svg>         6
  pages with <svg> inside them         0
  <image> elements inside <svg>        3
  spine media types                    image/svg+xml x6
```

Converted:

| build | images drawn |
|---|---|
| `main` | 0 |
| #157 | 0 |
| #159 | 0 |

## Mechanism

The document parses and is not filtered out — `converter.py` admits any media type containing `html` or `xml`, and `image/svg+xml` contains `xml`. It reaches `extract_blocks_from_html`, which walks from `<body>`. An SVG document has no `<body>`, so the walk finds nothing and returns an empty block list:

```
root tag: {http://www.w3.org/2000/svg}svg
<image> count in doc: 1
blocks extracted: 0
```

## Priority

Low, on current evidence. The reporter of #154 measured their own library at 0 of 226 books using the rooted form, against 104 of 226 using the nested form. No corpus book uses either. So this is a real gap with no known affected book — worth fixing when the SVG path is touched again rather than on its own.

Filed so the distinction is written down: #154's title reads as covering both forms, and a fix for the nested case can look complete while this stays broken.

kmnkv1:
Taking you up on the offer. Stacked on #159, since this calls the helper that PR adds — happy to fold it into #159 instead if you'd rather review the SVG path once; say the word and I'll squash it there.

The gap is bigger than the missing pictures, by either of two routes. Measured on the actual `IDPF/epub3-samples` book:

| build | images drawn | characters leaked into the body |
|---|---|---|
| `main` | 0 | 2,334 |
| this branch | 3 | 0 |

`main` does convert that book — three of its six pages "have text", and the text is the stylesheet and the RDF its drawing program left behind, reaching the reader as body content. A rooted SVG that is *clean* takes the other route: every page yields zero blocks, `spine_items_ordered` ends up empty and #72's guard fires with `ValueError: No spine items with extractable text`. So the failure is either silent garbage or a refused conversion, depending on how tidy the producer was.

Both leaks are closed here. Emitting image blocks alone was not enough: a page that draws no bitmap still fell through to the flat-text fallback, where `body.xpath("string()")` takes everything. An SVG document's text nodes are never reading content, so that fallback is skipped for them — the page contributes its `<image>` elements or nothing. SVG `<text>` is what that gives up, knowingly, and the comment says so.

The mechanism is exactly as you described it: the media-type test admits `image/svg+xml` because it contains `xml`, the document parses, and then `extract_blocks_from_html` walks from `<body>`, which falls through to the root element for an SVG document. The walk then iterates `<image>` children, and no branch of `_walk` claims them. `_emit_svg_blocks` already exists for the nested case, so the fix is to call it when the resolved `body` is itself an `<svg>`.

The sixth image is the tail page #157 restores, so this inherits the same ordering: #157 first, then #158, #159, this.

Four tests — one image, two images in one document, a rooted SVG that draws nothing, and the nested form unchanged as the control. The reproducer in #166 is `add_image_page(..., wrapper="svg-root")`, so once that lands this shape can be built without the sample book.

Closes #165.


jloutsch:
Gates are clean — 809 passed, tier3_strict 13 byte-identical, ruff clean — and the fix does what it says: on the `IDPF/epub3-samples` book from #165, images drawn go 0 to 3.

Two things from testing it against that sample rather than a reconstruction.

**The `ValueError` is real but narrower than the PR states.** The actual sample converts fine on `main`:

```
main: conversion SUCCEEDED   (3 sections, 0 images)
```

It survives because three of its six pages yield "text" — and that is the part worth looking at:

```
cover.svg          <image>=0 <text>=0  blocks=1  chars=1894
Tsuru_wiki-1b.svg  <image>=0 <text>=0  blocks=1  chars=220
```

Neither has a single `<text>` element. What reaches the reader is the stylesheet and the RDF metadata:

```
cover.svg          '.st0{fill:#AAB2AB;} .st1{fill:url(#path2455_2_);stroke:#000000;...'
Tsuru_wiki-1b.svg  '<metadata id="metadata2963"> <rdf:RDF> <cc:Work rdf:about="">...'
```

So the failure mode on `main` is either a hard `ValueError` when the SVG is clean, or silent garbage in the body when it is not. Your reconstruction hit the first; this book hits the second. Both are worse than "the pictures are missing", and your framing is right that it deserves more than low priority — it just gets there by a second route.

**This PR fixes one of the two leaks, not both.** After it:

```
Tsuru_wiki-1b.svg  blocks=0            <- metadata leak gone
cover.svg          blocks=1 chars=1894 <- CSS still reaches the reader
```

`cover.svg` has zero `<image>` elements and one `<style>`, so `_emit_svg_blocks` finds nothing to emit and the walk falls through to text extraction, which takes the stylesheet with it. `_is_non_rendered` only tests `hidden` and `epub:type`, and an SVG `<style>` or `<metadata>` carries neither.

A vector-art page that draws no bitmap is a normal thing for a cover to be, so this is not a corner. Worth handling here while the SVG path is open — either by treating `<style>`, `<metadata>`, `<defs>`, `<title>` and `<desc>` as non-rendered inside SVG, or by having a rooted SVG emit only image blocks and nothing else.

On the stack: yes to keeping it separate from #159. It is a different defect with a different reproducer, and #159 is already carrying two commits that are not its own. Land order stays #157, #158, #159, then this.

kmnkv1:
Both right, and the second one was a real hole in the fix. Fixed in a40837b.

You were correct that the `ValueError` framing was too narrow — I had reconstructed the shape rather than fetching the book, and my reconstruction's SVGs were clean, so it took the other route. On the actual sample:

```
                    images drawn   characters leaked into the body
  main                    0                 2,334
  first commit            3                 1,894
  with a40837b            3                     0
```

The leak survived because emitting image blocks is not enough: a page that draws no bitmap still produced no blocks, fell through to the flat-text fallback, and `body.xpath("string()")` takes the lot. `cover.svg` is exactly that page — vector paths and one `<style>`.

I went with your second option rather than the first. An SVG document's text nodes are never reading content — stylesheet, RDF, ids inside `<defs>` — so the fallback is skipped for SVG documents entirely, rather than enumerating `<style>`, `<metadata>`, `<defs>`, `<title>`, `<desc>` as non-rendered. The enumeration would have to stay complete against whatever a drawing program emits next; "an SVG page contributes its `<image>` elements or nothing" does not.

What that gives up is SVG `<text>`, and the comment says so out loud: nothing downstream renders it today, and the alternative on the same path is the stylesheet reaching the reader. Worth revisiting with a reference file that has some — the sample has none.

Four more tests: a stylesheet, metadata, a page that draws a bitmap alongside both of those, and an XHTML page still using the flat fallback as the control. PR body updated to the measured numbers.

Noted on the stack — kept separate, land order #157, #158, #159, then this.

kmnkv1:
Rebased onto the merge of #157, through #158 and #159. Gates green: 821 passed, tier3_strict 13 byte-identical, ruff clean. Land order is now just #158, #159, then this.

kmnkv1:
Rebased onto the merge of #166, through #158 and #159. Gates: 833 passed, tier3_strict 13 byte-identical, ruff clean.

Now that #166 is in, the reproducer for this is `add_image_page(..., wrapper="svg-root")` — a fixture can build the shape without the sample book.

jloutsch:
Rebase verified: 836 passed here.

**And the leak is gone.** Your comment does not mention it, so in case it was incidental to the rebase rather than deliberate — I re-ran the check from my earlier review against the sample:

```
                   before            after
cover.svg          1894 chars CSS    0 blocks
Tsuru_wiki-1b.svg  220 chars RDF     0 blocks
p002.svg           0 blocks          1 block, the image
```

A rooted `<svg>` with a `<style>` and no `<image>` used to put the stylesheet into the reader's body text. It no longer does. That was the half of this the earlier version left open, so as far as I am concerned the scope question on #165 is now fully answered.

Nothing outstanding here. It lands after #159, which still has the `<defs>` case open — same shape as what you just fixed, one level down. I have noted it there.

kmnkv1:
Re-stacked on #159 after the `<defs>` fix landed there — a rooted SVG gets the same rule, so a definition container in an .svg page is skipped too. Gates: 844 passed, tier3_strict 13 byte-identical, ruff clean.

kmnkv1:
Rebased onto the merge of #159, which carried the `<defs>` fix — so a rooted SVG gets it too. Gates: 851 passed, tier3_strict 13 byte-identical, ruff clean.

This is the next one in the chain.

jloutsch:
#159 is merged, and merging it broke this one. Squash-merge collapsed its two commits into one on `main`, so your branch still carries the originals and git now sees an add/add conflict in `converter.py` and `test_svg_images.py`. Our doing, not yours — and it will happen to every stacked PR we squash-merge.

The rebase is a one-liner, because the first two commits here are #159's and are already on `main`:

```
git fetch origin main
git rebase --onto origin/main 3a6a6c1
git push --force-with-lease
```

`3a6a6c1` is your "do not paint an `<image>` a definition container holds", the last commit that landed with #159. That replays only `d74860f` and `9d44b01` onto `main`.

I ran exactly that here to check it comes out clean, and it does — no conflicts, and verified against the sample rather than just the suite:

```
suite        854 passed
byte gate    13 byte-identical
ruff         clean

cover.svg    0 blocks, no CSS leak
p002.svg     1 block, draws its image
```

The result is on `origin/work/167-rebase` if you would rather cherry-pick than redo it.

I could have pushed it to your branch — `maintainerCanModify` is on — and decided not to. It is your repository, you mentioned being away for a day or two, and a force-push would take anything you have on that branch with it without asking. Not worth it for a mechanical conflict on a PR with nothing behind it.

No rush. #160 needs hardware regardless and #168 is unclaimed, so nothing is waiting on this.

kmnkv1:
Already done — our messages crossed by about two minutes. The branch has been sitting on `main` since 14:10, and the command was the same one you wrote, down to the base SHA:

```
git rebase --onto origin/main 3a6a6c1 fix/165-svg-root
```

Confirmed against yours rather than assumed: `origin/work/167-rebase` and this branch have **the identical tree**, so you can drop it. GitHub agrees the PR is clean:

```
mergeable = MERGEABLE, state = CLEAN, 2 commits, head 29dd4cd
```

Gates here match yours: 851 passed, tier3_strict 13 byte-identical, ruff clean.

Noted on the stacked-squash problem — it will hit #176 the same way when this merges, and I will re-run the same `--onto` then rather than making you write the SHA out again.

And thank you for not force-pushing to the branch. Nothing was on it, so it would have cost nothing this time, but the reasoning is the one I would want applied when something is.

jloutsch:
Crossed in the post — yours was up before mine went out, and identical trees is the right way to prove it rather than "should be the same". `origin/work/167-rebase` is deleted.

Confirmed at this head: 851 on your run, 854 on the tree I had, and #176 stacks on it cleanly at 866. Clean/mergeable.

Nothing outstanding. This goes in next.

PR #171, opened 2026-09-19
fix: make the tier-2 gap answerable instead of silent (#99)

jloutsch:
Addresses #99. **Stacked on #150** — it needs that pin refresh to have something truthful to record, so merge #150 first.

Tier 2 decodes kfxgen's output with jhowell's full `kfxlib`. That zip is not in this repository, so all 67 of its tests skip on every PR and have since the suite was added. A skip is not a failure, so the run goes green and reads as "everything checkable was checked" — which is how #91 came to cite tier-2 coverage that had not happened.

The zip cannot be committed, so tier 2 cannot become an ordinary CI gate. What *can* run in CI is a check on the record of the last real run.

## What changed

- **`scripts/tier2_record.py`** runs the suite locally and records the pin it ran against, the date and the count. It refuses to record a run that skipped anything or collected nothing, so a partial run cannot masquerade as a full one.
- **`tests/unit/test_tier2_record.py`** is tier 1 — no zip, no upstream import — so it runs on every PR exactly where tier 2 cannot. It fails when the vendored pin moves without a tier-2 run behind it. Refreshing the pin is when tier 2 has something to say and when it is easiest to skip.
- **CI** writes `Tier 2 did not run` into the job summary with the record inlined, rather than leaving 67 skips in a log nobody opens.

Deliberately **not** a hard failure. Tier 4 fails when unsigned because a human with a Kindle is supposed to act; nobody can act on this one, and a red run would charge a licensing constraint to contributors.

## The documentation was the actual problem

`CONTRIBUTING.md` asserted the cadence that does not happen — tier 2 as "Every PR" in the tier table, and again in the gate ladder. That is the mechanism behind #91's reasoning: nobody inferred tier-2 coverage from a green tick, they read a table that stated it. Both now say where tier 2 really runs.

## First record

```json
{ "upstream_pin": "20260827", "ran_on": "2026-09-19", "tests_passed": 67 }
```

Verified by mutation — a matching record passes; a moved pin, a deleted record and a zero count each fail with the procedure in the message.

## What this does not catch

A change to kfxgen's own generator that breaks decode compatibility without touching the pin. The record would still match. A passing record means "tier 2 was run against this upstream", never "kfxgen decodes correctly today". The test says so in its docstring.


kmnkv1:
Not a review — a data point for the premise, since I have a machine where tier 2 can actually run.

It runs, and it is green:

```
                              tier 2 (tests/integration/test_kfxlib_diff.py)
  origin/main                 66 passed, 1 failed
  main + my six open PRs      66 passed, 1 failed
```

The one failure is `test_vendored_pin_matches_the_zip_it_describes`, and it fails for the right reason — my `kfxlib` is 20260824 against the committed pin of 20260520. The 66 that matter, the differential decode itself, pass identically on both, which also says my changes do not move decode behaviour.

Two caveats on what that is worth. My `kfxlib` came out of **KFX Output**, not KFX Input — both ship the library, but it is not the same zip your record names. And 20260824 is neither the current pin nor the 20260827 that #150 moves to, so this is adjacent evidence rather than a record you could adopt.

What it does support is the shape of this PR: the suite is runnable, it is the *zip* that cannot be committed, and a record of where it last really ran is the honest thing CI can check. If a second machine's run is ever useful — on a pin refresh, say — I can run it and report the numbers; I cannot produce a record you should trust, since mine is the wrong plugin.

Unrelated but adjacent: #173 is the Windows side of the same "a gate that cannot gate" family — two assertions in `test_safe_write` that encode POSIX semantics. The defense they cover does hold on Windows; the assertions cannot.

jloutsch:
This is the most useful thing anyone could have posted here, and the caveat is the useful part.

Your numbers match mine exactly — I get the same 66 passed / 1 failed on `main` with a mismatched zip, and the same one failing test for the same reason. Two machines, two different archives, identical differential-decode results is a stronger statement about this PR's premise than either of us could make alone: the suite runs, it is green, and it is the *zip* that cannot be committed rather than anything being broken.

**The KFX Output point is a hole in my mechanism, and I have closed it.** You were right to flag it and right that your record is not one to adopt, but the reason generalises past your machine: `upstream_pin` was copied from the sidecar, which makes it a claim about a file rather than about a run. `test_vendored_pin_matches_the_zip_it_describes` compares the sidecar against `kfxlib/version.py` inside the archive, so a *version* mismatch already blocks a record — but two plugins shipping the same library at the same version pass that check identically. A record could have been written naming an upstream that was never the one tested.

The record now reads the version out of the archive under test and carries a digest of the archive itself:

```json
{
  "upstream_pin": "20260827",
  "kfxlib_version": "20260827",
  "zip_sha256_prefix": "338809c18e5f9bb7",
  "ran_on": "2026-09-19",
  "tests_passed": 67
}
```

Same version, different bytes, different record. `test_record_says_which_upstream_was_actually_read` keeps it that way.

That also answers your offer better than a yes would have. A second machine's run is worth having on a pin refresh, and with the digest in the record the provenance question answers itself — if your archive ever *is* the same bytes, that is now visible rather than assumed.

Thank you for #173 as well. "The defense holds, the assertions do not" is the outcome I was hoping for and not the one I expected, and measuring it rather than reasoning about it is what made the difference. I will pick it up separately.

One note on this PR's own state: it targets `chore/refresh-vendored-kfxlib-20260827`, and the workflow only triggers on `pull_request` into `main` or `develop`, so it has never had a CI run of its own. Which is, with some irony, exactly the failure mode it exists to make loud. #150 lands first and then this retargets to `main` and gets one.

PR #174, opened 2026-09-19
test: one builder for test JPEGs

jloutsch:
Three implementations of "make a JPEG that declares its size" had accumulated: `jpeg_of` in `tests/_helpers`, and hand-rolled `_jpeg` copies in `test_image_optimize.py` and `test_reproducible_output.py`. Two more were arriving with the open PRs.

Divergence between them is not academic. `MINIMAL_JPEG` has been silently size-less since it was written, and #163 was a cover golden that could not see a cover change because of it — a change that moved 90 of 90 corpus books. Part of why that stayed invisible is that there was no single place where "how we build a test JPEG" was decided.

## What changed

Both local copies now alias `jpeg_of`, which arrived in #166 with the shape builders that motivate it. One builder in the repo:

```
tests/_helpers.py:43:def jpeg_of(width: int, height: int) -> bytes:
```

No behaviour change. **805 passed, all 13 goldens byte-identical** — nothing feeding a golden moved.

Also corrects a comment in `test_reproducible_output.py` claiming `MINIMAL_JPEG` "has no SOF at all". It has one, at offset 138; a length-following parser never reaches it. That misreading is the same one that made #163 hard to see, sitting in the tree as documentation.

## Why `SIZED_JPEG` is still fixed bytes

It feeds the `with_cover` golden. Deriving a golden's *input* from a helper couples the committed output to that helper — editing `jpeg_of` for some unrelated fixture would move `with_cover.kfx` and fail the byte gate far from the change. Golden inputs hold still on purpose; `jpeg_of` is for fixtures built at run time. That reasoning is now in the file.

## Deliberately not included

A structural guard asserting no module defines its own `_jpeg` would prevent this recurring, and this repo's habit is that every gate gets a gate. It is not here because **#158 and #160 each still add a copy**, and landing the guard now would turn both red for a reason unrelated to their content — which is the wrong way to treat a PR already under review. Worth adding straight after those two land.


jloutsch:
Updating the reasoning in this PR's description: the contributor has just dropped the `_jpeg` copies from #158, #159, #160 and #167 in favour of `jpeg_of`, so the thing I was avoiding no longer exists.

```
files defining _jpeg, after their rebases
  origin/main   test_reproducible_output.py, test_image_optimize.py
  pr158         test_reproducible_output.py, test_image_optimize.py
  pr160         test_reproducible_output.py, test_image_optimize.py
```

Only the two pre-existing ones, which this PR removes. So once this lands there is exactly one builder, and **the structural guard I deferred can go in without turning anyone's PR red** — which was the entire reason for leaving it out. Worth adding as a follow-up right after this merges.

PR #176, opened 2026-09-19
fix: draw a page the stylesheet paints as a background (#168)

### LINKED ISSUE #168: Pages drawn with a CSS background-image lose every picture — six books in a 226-EPUB library convert with no art at all
Found while measuring #152 across a 226-EPUB library of illustrated children's books. Six of those books reach the reader with **no pictures at all** — not a page dropped here and there, every image in the book.

## What happens

The page holds no ``. The picture is a CSS `background-image` on an empty `<div>`:

```html
<body class="calibre">
  <div class="page" id="piv"><div class="pimg5" id="img_iv"></div></div>
</body>
```

```css
.pimg5 { background-image: url(images/00004.jpeg); background-repeat: no-repeat;
         display: block; height: 1280px; position: absolute; width: 800px }
```

`extract_blocks_from_html` emits an IMG token for ``, and (with #159) for `<svg><image>`. It has no path for this, so the page yields a block of no text and no token. `extract_images_from_oeb` still emits the `$164`/`$417` resources from the manifest — and #102 then prunes them again, correctly, because no `$259` entry displays them.

Measured, converting each book on `main`:

```
  spine docs   background images   result
        227          226           218 chapters,  0 image tokens
        227          226           218 chapters,  0 image tokens
        227          226           218 chapters,  0 image tokens
        226          225             5 chapters,  0 image tokens
        226          225           222 chapters,  0 image tokens
        101          100            21 chapters,  0 image tokens
```

1,228 page images across six books, none of which reach the KFX. The books still convert and still carry their text, which is why this is silent: nothing fails, the pages are simply blank where the art was.

## Scope

Six of 226 books here, so roughly 3%. All six are the same producer's output — a print-to-EPUB chain that positions each scanned page as a background layer with the text absolutely positioned over it. None of the 90 Gutenberg corpus books does this, and neither do any of the fixed-layout samples in #152's data; it seems specific to page-scan books from that toolchain.

## What a fix would involve

More than the SVG case, and I'd rather flag it than guess at the shape:

* the reference is in CSS, not markup, so it needs the Stylizer — which exists only inside Calibre, so the unit tests would need the injected-resolver pattern from #9;
* `background-image` is not in the property list `_build_style_resolver` requests, and its computed value is a `url(...)` that has to be resolved against the stylesheet's own path, not the document's;
* one `<div>` can legitimately carry a decorative background that is not content, so emitting a token for every one of them would be wrong — the six books here are recognisable by the page being *nothing but* the div, which is a narrower rule than "any background-image".

Happy to take it if you want it, but it is a bigger change than the four I have open and I don't have a second producer's books to check the rule against. Filing it so the gap is recorded either way.

kmnkv1:
Taking it. Stacked on #167, as you suggested — the walk is the same code.

Your three-part rule, with your third part included, and it is the third that earns its place: `background-repeat: no-repeat` is what separates a page scan from a texture, and every producer with this shape sets it.

Two things from your comment that changed the implementation:

**The path really is irrelevant**, and I checked it on a real book rather than taking it on faith: the `url(...)` values resolve against the OEB manifest by basename, 6 of 7 pages on the first book I tried. The seventh is the cover, which is handled separately anyway.

**The check had to go before the leaf/container split in `_walk`.** My first attempt put it in the container branch and it never fired once: such a div is a leaf whose text is empty, so the leaf branch drops it before the container branch is reached. Worth knowing if you ever move it.

## The measurement you asked for

Rule on against rule off, on the same OEB, so nothing else can differ. 226 books, no errors:

```
  books whose outcome changes            6
    gaining images                       6
    losing anything                      0
  images across the library         18,664 -> 19,796   (+1,132)

  no pictures at all before    0 ->  226   page-scan book  (x3)
                               0 ->    5   a text book whose front matter is scanned
  only a cover before          1 ->  226   (x2)
```

**Books that gain something that is not a page: 0.** I checked every firing element by its box rather than assuming: 605x880 and 756x1098, full-page scans throughout. The `0 -> 5` is the one that could have been decoration and is not — four scanned front-matter pages and a scanned back page in a book that is otherwise ordinary text and ``. That book is also the argument for not narrowing the rule further to "the whole book is this shape".

## One structural repair, named because it is a large diff for one rule

One book goes from 222 chapters to 4. Its pages used to yield nothing, so they never entered `spine_items_ordered`, so the publisher's own nav coordinates could not resolve and the book was sliced into invented filename chapters instead:

```
  before   222 chapters, 17,964 words, 220 titled `part0197`-style and omitted from the nav
  after      4 chapters, 18,185 words, the publisher's own four labels, none omitted
```

The extra 221 words are pages that used to be dropped outright.

## What it does not do

Off outside Calibre — the reference is in CSS, so it needs the Stylizer, and `_build_style_resolver` returns None without it. The twelve tests use the injected-resolver pattern from #9: six that a page is drawn (including quoted `url()`, the box size, and text over the scan surviving beside it), six that decoration is not (tiled, unstated repeat, a box with text, a box with children, `none`, and no resolver at all).

Gates: 856 passed, tier3_strict 13 byte-identical, ruff clean.

Closes #168.


kmnkv1:
Correcting my own numbers before you spend review time on them. The measurement above counts **tokens**, and for four of the six books those tokens do not become pictures. End to end, through `ebook-convert` with the plugin installed:

```
                        tokens   image entries in the KFX   why
  DWK-style page scan      225            225               all JPEG
  a second page-scan book    5              2               see below
  three page-scan books    226 each         2 each          images are GIF
  one more                 226              3               images are GIF
```

**Four of the six are blocked by something else entirely.** Their page scans are GIF, and `extract_images_from_oeb` accepts only JPEG and PNG:

```
  Skipping image 'images/00008.gif': unsupported format (magic bytes 47494638);
  only JPEG and PNG are emitted as KFX resources
```

225 of those per book. It is logged, not silent, and it predates this PR — but it means this change alone does not give those readers their pictures. Worth noting `$286` is GIF in the format table and upstream kfxlib lists it among the fixed-layout image formats, so it may be a real gap rather than a deliberate limit; I will file it separately rather than widen this PR.

**The `0 -> 5` book is not what I said it was.** I called it "a text book whose front matter is scanned". It is a page-scan book like the others — 100 elements carry a background url — and the rule fires on 5 of them. The other 95 have the overlay text **inside** the background div rather than as a sibling, so condition (2) excludes them:

```html
<div id="page051" class="calibre53">          <- background: the scanned page
  <div class="cls_191">Worm's body glowed</div>   <- overlay text, a child
```

That is the rule being conservative exactly where you asked it to be, and I would rather leave it there than widen (2) to "no text of its own, children allowed", which would fire on every decorated wrapper in the corpus. But it means the shape has a second variant this does not reach, and you should know that before judging the rule's coverage.

**One thread I have not closed.** In that book, 4 body tokens survive chapter assembly and only 2 reach the file as `$164` resources. The missing two are on the Title Page chapter. I have not found where they go yet — it is downstream of this change and reproduces without it, so I do not think it belongs to this PR, but I am not certain enough to say that flatly. I will keep pulling on it.

So the honest claim for this PR is narrower than the one I posted: **it is correct on every book it fires on, it adds nothing that is not a page, and on this library exactly one book gets its 225 pictures out of it today.** The rest wait on GIF support, or on the second variant of the shape.

kmnkv1:
Rebased onto the merge of #159, through #167. Gates: 863 passed, tier3_strict 13 byte-identical, ruff clean.

jloutsch:
The correction is the most valuable thing in this PR, and I want to be specific about why: you caught a **token-versus-resource** conflation in your own headline number, before anyone spent review time on it. That is the exact mistake I made earlier in this repo's history — counting what the converter emits rather than what reaches the file — and it is not a mistake that surfaces on its own.

**Verified independently, not taken on trust.** Gates: 866 passed here, tier3_strict 13 byte-identical, ruff clean. And the rule, reproduced from scratch rather than by reading your tests:

```
page scan (empty div, no-repeat)   -> images/00001.jpeg
tiled texture                      -> no token
repeat unstated                    -> no token
box with text of its own           -> no token
box with children                  -> no token
no resolver at all                 -> no token
```

Six for six. `background-repeat` earning its place is the part I would not have been confident about without the library behind it.

**The GIF finding is real and better evidenced than you put it.** The repo's own comment names the symbol and the next line declines to use it:

```
native_generator.py:539   # $285 = JPEG, $284 = PNG, $286 = GIF
native_generator.py:540   format_sym = "$285" if image_format == "jpeg" else "$284"
```

So GIF is known to the format table and never emitted, and `extract_images_from_oeb` filters it before the generator would ever see it. That is a gap, not a decision, on the evidence available — please do file it. Four of six books here, and it is logged rather than silent, which is the one mercy.

**The `0 -> 5` book.** Your instinct is right and I would not widen condition (2). "No text of its own, children allowed" fires on every decorated wrapper in a library, and the cost of being wrong runs the expensive way: a spurious picture in the reading flow is worse than a missing one, because the reader cannot tell it was not meant to be there. Leave the second variant unreached and record it.

**The unclosed thread — 4 tokens, 2 resources.** Leaving it open and saying you are not certain it belongs to this PR is the right call. It reproduces without this change, which is the test that matters. If it helps: `#102` prunes any `$164` nothing references, so two tokens failing to resolve *upstream* of the generator would look exactly like this. `_emit_text_chunks` drops a token whose basename is not in `image_resources` silently — that is where I would look first, and a Title Page chapter is a plausible place for a basename to collide or for the cover path to intercept it.

## On merging

Clean and mergeable. It goes after #167. The honest framing for the release notes is yours, not the original one: correct on every book it fires on, adds nothing that is not a page, and one book in this library gets its pictures from it today.

kmnkv1:
Thread closed, and your hint was the right place to look — though the loss turned out to be one step earlier than `_emit_text_chunks`.

It is `_replace_title_page`. The Title Page chapter carried those two tokens, and the replacement overwrites the body with title + author:

```
without metadata   Title Page = '<IMG images/00020.jpeg><IMG images/00021.jpeg>'
with metadata      Title Page = 'Vortex\n\nby\n\nTracey West'
```

So the basenames never reach `image_resources` lookup at all — the tokens are gone before the generator sees them. Reproduces on `main` without this PR, as suspected.

Filed as #178 with the library measurement: **181 of 226 books, 314 images**, counting `preserved_images` as kept so the #117 path does not inflate it. Same shape as #117 one page earlier in the book, and the fix is the same key the generator already emits.

GIF is #177, filed before your comment — measurement, the two lines from `native_generator.py`, and the `FIXED_LAYOUT_IMAGE_FORMATS` evidence that $286 is nameable. I offered to build a GIF-versus-PNG pair and settle the render question on the Colorsoft along with #160's pass.

Thank you for the merge, and for the framing note — I will use the narrower wording if release notes want a line.

jloutsch:
Confirmed #178 here before reading your diagnosis, and you are right that it is one step earlier than I guessed:

```
without metadata   tokens=2   <IMG images/00020.jpeg>  <IMG images/00021.jpeg>
with metadata      tokens=0   'A Title\n\nby\n\nAn Author'
```

`_replace_title_page` overwrites the body wholesale, so the basenames never reach a lookup. 181 of 226 books is a bigger number than I would have guessed from two missing pictures.

**One request, and it is about the repository rather than about the work.** That comment names a real title and author in its output. Could you delete and repost it with those two values replaced — `'A Title\n\nby\n\nAn Author'` reads identically for the purpose?

Not a rule I had written down anywhere, so this is me telling you late rather than you missing something. The reasoning:

- This repository is public and permanently indexed. A title and author sitting next to "a 226-EPUB library" invites inferences about how such a library was assembled, and those inferences land on you, not on the project. I would rather nobody has to make them.
- The README describes its own corpus by shape for the same reason, and `research/describe_epub.py` prints no title, author, identifier or filename precisely so that a book can be characterised in a public thread without being named. You have been using it that way all day — this is the one place it slipped, in a debugging paste rather than a measurement.

**Delete and repost rather than edit.** GitHub keeps the previous revision in a comment's edit history, so editing leaves the original text one click away; deleting removes it.

For future reference, the pattern that has worked well in this repo: describe a book by the shape that matters — "a 227-page scan book", "a text book whose front matter is scanned", "one book with 443 ids inside table cells". Every number you have reported so far has been legible without a title attached, which is why this is the only instance.

Nothing else outstanding. #177 and #178 are both good filings, and yes please to the GIF-versus-PNG pair on the Colorsoft alongside #160 — that settles two questions in one session.

kmnkv1:
Thread closed, and your hint was the right place to look — though the loss turned out to be one step earlier than `_emit_text_chunks`.

It is `_replace_title_page`. The Title Page chapter carried those two tokens, and the replacement overwrites the body with title + author:

```
without metadata   Title Page = '<IMG images/00020.jpeg><IMG images/00021.jpeg>'
with metadata      Title Page = 'A Title

by

An Author'
```

So the basenames never reach `image_resources` lookup at all — the tokens are gone before the generator sees them. Reproduces on `main` without this PR, as suspected.

Filed as #178 with the library measurement: **181 of 226 books, 314 images**, counting `preserved_images` as kept so the #117 path does not inflate it. Same shape as #117 one page earlier in the book, and the fix is the same key the generator already emits.

GIF is #177, filed before your comment — measurement, the two lines from `native_generator.py`, and the `FIXED_LAYOUT_IMAGE_FORMATS` evidence that $286 is nameable. I offered to build a GIF-versus-PNG pair and settle the render question on the Colorsoft along with #160's pass.

Thank you for the merge, and for the framing note — I will use the narrower wording if release notes want a line.

kmnkv1:
Done — deleted and reposted rather than edited, so the original is not in an edit history.

#178 carried the same two values in its body and is redacted too, but with a caveat you should know rather than assume: an issue **body** cannot be delete-and-reposted the way a comment can. Editing it removes the text from view and leaves it in the body's edit history, which is the exact exposure you described. Refiling would leave the original issue behind with the text still in it, and I cannot delete an issue in your repository. So if you want it gone entirely, deleting #178 and letting me refile from the redacted text is the only route, and I am happy to do that — say the word. Otherwise it is out of view and the history is a click away for anyone who thinks to look.

You are right that it was a debugging paste rather than a measurement, and that is the honest description of how it happened: I had been describing books by shape all day because shape was what the argument needed, and then pasted a variable's value verbatim because I was chasing a mechanism rather than reporting a result. The rule is a good one and the reasoning for it is better than "titles are sensitive" — thank you for writing it down rather than just asking.

Colorsoft pass: #160's three checks plus the GIF-versus-PNG pair for #177, one session, a day or two out.

kmnkv1:
Correction to my own "both places": I swept every thread I have posted in and found a third, weaker instance — #154's body named a series (no title, no author) as the source of a markup example. Redacted to "the `cover.xhtml` of a page-scan children's book", which is what the sentence needed anyway.

Everything else is clean: I checked the bodies and comments of all ten issues and ten pull requests I have touched against the series and authors in that library. Three instances total, all mine, all in prose rather than in measurements — which matches your read that the measurements were never the problem.

Issue #177, opened 2026-09-19
Page scans stored as GIF are dropped — four books in a 226-EPUB library convert with no pictures


kmnkv1:
Found while measuring #176, and split out of it as promised: it is the reason that PR helps fewer books than its own numbers suggest.

`extract_images_from_oeb` accepts JPEG and PNG and nothing else:

```python
if data[:3] != b"\xff\xd8\xff" and data[:4] != b"\x89PNG":
    skipped_unsupported += 1
    log.warn(f"  Skipping image {href!r}: unsupported format ...")
```

It is logged rather than silent, which is right, and the generator's own comment already names the missing case:

```python
# $285 = JPEG, $284 = PNG, $286 = GIF
format_sym = "$285" if image_format == "jpeg" else "$284"
```

## What it costs

Four books in a 226-EPUB library of illustrated children's books are page scans stored as GIF. Converting them today:

```
  DWK-style page-scan book   225 images skipped as unsupported,   2 reach the KFX
  second                     225                                   2
  third                      225                                   2
  fourth                     224                                   3
```

Every page of each. The books convert, the text is there, and the pictures are not — the same "silent where something else survives" shape as #168 and #165, one layer down.

Outside that shape it is narrower than it looks: of 226 books, those four are the only ones where GIF carries the content. A stray decorative GIF elsewhere loses an ornament, not a page.

## Is GIF actually emittable

I have not verified on hardware, so this is the paper case only:

* `$286` is GIF in the format table the generator already comments;
* upstream kfxlib lists it as a fixed-layout image format — `FIXED_LAYOUT_IMAGE_FORMATS = {"$286", "$285", "$548", "$565", "$284"}` in `yj_structure.py`;
* so the container can name it. Whether a Kindle renders it from a sideloaded KFX is the open question, and the only way to answer it is a device.

There is a second question behind it: an animated GIF has no meaning here, and a page scan stored as GIF is a still. If GIF turns out not to render, converting to PNG on the way in would reach the same books — at a size cost, and PNG of a scanned page is large.

## What I would want before writing it

A device check that a KFX carrying `$161: $286` renders at all. I have a Colorsoft and can build the pair — one book with the image passed through as GIF, one with it converted to PNG — and report which of them shows a picture. Say the word and it goes on the same pass as #160's.

Not proposing a fix in the dark; the measurement above is the part worth having recorded either way.

jloutsch:
Written as #183, and deliberately not the fix you declined to write.

[...]

Issue #173, opened 2026-09-19
test_safe_write: two assertions are POSIX-only and fail on Windows — the defense itself holds


kmnkv1:
Asked for in [#166](https://github.com/jloutsch/kfxgen/pull/166) — you wanted these as their own issue rather than as an aside, because a genuine Windows failure would mean the symlink defense is weaker where nobody tests. Measured: **it is not. Both failures are POSIX assumptions in the tests, and the behaviour under test is correct on Windows.**

Python 3.13.1, Windows 11, `pytest tests/unit/test_safe_write.py` → 20 passed, 2 failed, 1 skipped.

## 1. `TestSafeWriteHappyPath::test_mode_is_0644`

```
assert (target.stat().st_mode & 0o777) == (0o644 & ~_current_umask())
E   assert (33206 & 511) == (420 & ~0)
```

`st_mode` is `0o100666`, so the masked value is `0o666`. Windows has no POSIX permission bits — `os.chmod` toggles the read-only attribute and nothing else, and there is no umask. The `0o644` the production code requests is not expressible, so the assertion cannot hold on this platform. Nothing is wrong with the write.

## 2. `TestSafeWriteSymlinkRejection::test_refuses_symlink_at_destination`

The defense assertions all pass:

```
OSError raised, errno == ELOOP (10062 on Windows)     ok
sensitive.read_bytes() == b"DO NOT OVERWRITE"         ok   <- not written through
symlink.is_symlink()                                  ok
os.readlink(str(symlink)) == str(sensitive)           FAILS
E   assert '\\?\C:\U...sensitive.txt' == 'C:\Users\K...sensitive.txt'
```

Only the last one, and only because `os.readlink` returns the extended-length `\?\` form on Windows. The symlink was refused, the target was not overwritten, the file is still a symlink — the defense did exactly what it exists to do.

So `O_NOFOLLOW` degrading to `0` does not leave the destination unprotected here: the `islink()` check is what rejects, and it fires. The TOCTOU window the docstring already describes is unchanged, and is the thing that is genuinely weaker without `O_NOFOLLOW` — but that is about the tmp slot, not the destination, and it is in-scope-declined by SECURITY.md.

## Suggested

Two small test changes, no production change:

* gate the mode assertion on `os.name != "nt"`, or assert "readable and writable by the owner" instead of an exact mode;
[...]

Issue #164, opened 2026-09-19
No golden exercises a computed image width, so the #145 rule and #96 ordering are untested


jloutsch:
Four goldens carry an `s_img_w0` image style, so the `sized` bucket looks covered. Every
one of them is `$56 = 100.0`:

```
body_images            s_img_w0  $56 = 100.0
captioned_images       s_img_w0  $56 = 100.0
unused_manifest_image  s_img_w0  $56 = 100.0
with_cover             s_img_w0  $56 = 100.0
```

100.0 is the fall-through, not a computed result. Every fixture image has unreadable
dimensions (see the `MINIMAL_JPEG` DQT defect), so `_classify` returns `sized` by way of
`not (w and h)` and the width comes from the `else 100.0` default. `_image_width_pct` is
never called.

Two guards are therefore not guarding.

## `_image_width_pct` (#145)

The rule is `min(100, px / 496 * 100)`. The interesting half is an image narrower than
the text column, which produces a width below 100. No golden can produce one, so a
regression in that arithmetic passes the byte gate untouched.

This is not hypothetical. The defect #145 fixed gave 2,462 of 4,436 corpus images a fixed
9.626% width, across 29 of 77 books and 10 covers, and it shipped. The golden corpus was
green throughout, and would be green again for the same class of error.

## Sorted style allocation (#96)

`s_img_w0`, `s_img_w1`, … are allocated in sorted width order because iterating the set
directly randomized symbol ids between runs, shifting every local symbol numbered after
them. Every golden has exactly one width bucket, so there is no ordering to get wrong.
The fixture set cannot distinguish sorted allocation from unsorted.

## Fix

Give the body-image fixtures images with real, differing dimensions — at least one
narrower than the 496px column so a computed width appears, and at least two distinct
widths in one book so the allocation order means something. Regenerate the affected
goldens.

This is a deliberate change with real churn: the same fixtures currently sit in the
unknown-size bucket, and moving them will alter their goldens. Worth doing as its own
piece of work rather than folded into an unrelated fix.

A useful shape assertion to add alongside: at least one golden must carry a width style
strictly between 0 and 100, and at least one must carry two distinct width styles. Both
fail today.

Issue #178, opened 2026-09-19
The title-page replacement discards the pictures printed on that page — 181 books, 314 images


kmnkv1:
The thread I left open on #176 — 4 image tokens surviving chapter assembly, 2 reaching the file — closes here, and it is not about that PR at all. It reproduces on `main` without it, and it is the same shape as #117 one page earlier in the book.

## Mechanism

`_replace_title_page` overwrites a title page's body with title + author:

```
without metadata   Title Page text = '\x00IMG\x01images/00020.jpeg\x01\x01w=756px\x00\n\n\x00IMG…'
with metadata      Title Page text = 'A Title

by

An Author'
```

Whatever the publisher printed on that page goes with it. In a page-scan book that is the scanned title page; in an ordinary illustrated book it is the vignette or the series logo above the title.

#117 made exactly this call for the contents page and kept the pictures — `_rebuild_contents_page` collects the tokens into `preserved_images` and the generator emits them. The title-page path has no equivalent.

## Measurement

226-EPUB library of illustrated children's books, chapters extracted with and without metadata so only this function differs. Images carried in `preserved_images` are counted as kept, so the #117 path does not inflate the number:

```
  books losing images   181 of 226
  images lost                  314

  by the chapter they were printed on
     281   'Title Page'
      20   'Cover'
       8   'TITLE PAGE'
       2   'Half Title'
       2   'Title'
       1   'Halftitle Page'
```

Between one and seven per book; the distribution is 1:19, 2:57, 3:93, 4:24, 5:7, 7:1.

The 20 on `'Cover'` are worth a second look by someone who knows that path — `TITLE_PAGE_TITLES` is `{"title page", "title"}`, so a chapter titled `Cover` should not be matching it, and I have not chased why those lose anything.

## Suggested

The same treatment #117 gave the contents page: collect the tokens before overwriting and hand them to the generator as `preserved_images`, which already knows how to emit that key (`native_generator.py:3010`). The replacement text is the point of the feature and should stay; the pictures are not what it exists to remove.

Worth deciding deliberately for one case: a title page that is *only* a scanned image, where the replacement produces "Title / by / Author" as text over — or instead of — a picture that already says exactly that. Keeping both is defensible, and so is preferring the text on a reflowable device. I would keep both, on the grounds that the scan is what the publisher set.

Happy to write it if you want it; it is small, and the fixture shape is already in `EpubBuilder`.

jloutsch:
[...]

PR #179, opened 2026-09-19
release: v5.8.0 — books made of pictures

jloutsch:
Fifteen commits since 5.7.4. Version to `5.8.0` — minor rather than patch, because this adds capability that did not exist: kfxgen now reads pages drawn as `<svg>`, as an SVG document, and as a stylesheet background, and ships pages the navigation never listed.

**Do not merge this without pushing the `v5.8.0` tag afterwards.** A bumped version with no tag means no GitHub Release.

## Why one release rather than seven

Every user-visible fix here is the same defect wearing a different hat: a page whose content is a picture rather than a paragraph reached the reader empty. They are separate code paths and one story, and the entry is written that way.

## What this release is really a record of

None of it was findable from this repository. The 90-book corpus is prose — not one of its books is fixed-layout, wraps art in `<svg>`, states a size on an image, or paints a page from a stylesheet. All seven user-visible fixes came from an outside contributor measuring against a 226-EPUB library of illustrated children's books.

Every number in the entry was reproduced here before being written down, including the ones that made their own PRs look worse. Two worth keeping:

- #159 alone was a **net regression** — 225 of 226 images lost on a real shape — which they found and reported on their own pull request.
- #176's headline number counted tokens rather than resources, which they corrected before review, reducing their own claim from six books to one.

## Not in this release

**#160** rewrites the cover section for every book and is the one change needing a device pass. It is independent, so holding fifteen commits for it would have meant the rest waiting on hardware availability. It goes in the next one.

## Known gaps, recorded rather than fixed

#177 (GIF page scans dropped), #164 (no golden exercises a computed width), #173 (two POSIX-only assertions), and the background-image rule's unreached second variant. All are in the entry rather than left for someone to rediscover.

## Verification

```
suite        866 passed, 67 skipped
byte gate    13 byte-identical
ruff         check + format clean
build        dist/kfxgen-plugin-5.8.0.zip
version      single source reads 5.8.0
```


kmnkv1:
Read it against my own measurements, since the entry says they were reproduced. Three corrections, all small, and one of them is mine to own.

**1. "one of them shipping 3 of its 7 pages" (#152) is not one of the 226.** That was a synthetic seven-page fixed-layout EPUB I built to demonstrate the shape — it is in the issue as a test EPUB, and the sentence as written reads as a real book from the library. The strongest real one I measured is 104 spine documents with 22 nav entries, where seven pages were dropped; the library-wide figures beside it (200 pages, 64 books) are the ones from real books.

**2. "Six books converted with no pictures at all" (#168) is four.** Of the six the rule changes, four had zero images and two had exactly one — their cover, nothing else. "Six books converted with no pages at all, four of them with no picture of any kind" would be exact, if that is not too fussy for a changelog.

**3. #178 is missing from Known gaps**, and I filed it after you had written this, so no wonder. The title-page replacement discards whatever the publisher printed on that page: **181 of 226 books, 314 images**, counting `preserved_images` as kept so the #117 path does not inflate it. It is the same shape as #117 one page earlier in the book. It is also where the unclosed thread from #176 went, so it belongs in this release's record of what is known and not fixed.

Everything else I can check matches: +277/+166 and the 58 single-token losses, 6,606/6,098/490 across 102 books, 104 of 226 wrapping art in `<svg>`, 1,894 characters of CSS, 222 chapters to four, 90 of 90 books changing behind a passing gate, and the 67/114/49 arithmetic.

On the attribution — thank you, and the two lines about #159 being a net regression and #176 counting the wrong thing are the right ones to keep. Both were caught because you kept asking for measurements rather than accepting arguments, which is worth more to the outcome than either finding.

**#160 status for your planning:** the device pass is a day or two out here. If it slips or your hardware frees up first, take it — it is the only thing between that PR and merge, and 5.8.1 need not wait on me.

PR #180, opened 2026-09-19
docs: say how to describe a book without naming it

jloutsch:
A contributor named a real title and author in a debugging paste on #176. When I asked them to redact it, I had to admit the rule was written down nowhere — so this writes it down.

The reasoning is not squeamishness about titles. Most defects here are shape defects: whether a page survives depends on how the navigation lists it and how its art is wrapped, not on what the book says. **The shape is the reproducer and the title is not.** What a title does add is an author and a work sitting next to a description of someone's private library, in a public repository that is permanently indexed — and the inference that invites lands on the contributor, not on the project.

The section points at `research/describe_epub.py` rather than asking people to invent a format, since it already prints exactly this and deliberately prints nothing else.

Two details included because neither is obvious:

- **Delete and repost, do not edit.** GitHub keeps the previous revision in a comment's edit history, so editing leaves the original one click away.
- **Gutenberg ids stay fine** (`pg11`, `pg3011`) — public domain, listed in the README, and what the tests key on. Without saying so the rule reads as banning them from test parameters.

Docs only. 866 passed, unchanged.


kmnkv1:
This is about my slip, so: the rule is right and the reasoning is better than the usual one. Both places are redacted — the #176 comment deleted and reposted, #178's body edited.

**One wrinkle worth a line in the section**, found while complying with it: "delete and repost" works for a *comment*, but an issue or PR **body** has no equivalent. Editing it leaves the original in the body's edit history, which is the exposure the paragraph above warns about; refiling leaves the original issue in place; and a contributor cannot delete an issue in someone else's repository. So for a body the only complete remedy is the maintainer deleting the issue and the contributor refiling from redacted text.

Suggested addition after the delete-and-repost bullet:

> This works for comments. An issue or pull-request **body** has no delete-and-repost: editing it leaves the original in the body's edit history, and only a maintainer can delete the issue outright. If a body needs redacting, edit it and say so — a maintainer can then delete and let you refile if the history matters.

The Gutenberg-ids carve-out is worth keeping for a second reason beyond public domain: it is what makes a corpus measurement checkable by someone else. "pg12082 was losing 97% of its links" is a claim a reader can verify; the same sentence about an unnamed book is one they have to take on faith. The line between the two is roughly "a book anyone can fetch" rather than "a book that is public domain", and the section reads correctly on that already.

